PR #1, opened 2019-03-26
Assignmentdev

zptro:
The functionality of main function assign() remains, but internally the assignment class is largely restructured.

### DIFF
--- a/Scripts/assignment.py
+++ b/Scripts/assignment.py
@@ -51,16 +51,15 @@ class AssignmentModel:
         """Assign cars, bikes and transit for one time period."""
         self.logger.info("Assignment starts...")
         self.set_matrices(matrices)
-        function_file = os.path.join(self.path,
-                                     "d411_pituusriippuvaiset_HM30.in")
+        function_file = os.path.join(self.path, param.func_car)
         process = self.emme_modeller.tool(
             "inro.emme.data.function.function_transaction")
         process(function_file)
         self.calc_road_cost(scen_id)
         self.assign_cars(scen_id, param.stopping_criteria_coarse)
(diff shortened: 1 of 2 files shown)


Scripts/assignment.py
```diff
@@ -365,6 +335,78 @@ def assign_transit(self, scen_id):
...
+                "actual_aux_transit_times": mtx["transit"]["aux_time"]["id"],
+            },
+        }
+        matrix_results(result_spec, scenario)
         self.logger.info("Transit assignment performed for scenario " 
-                        + str(scen_id))
\ No newline at end of file
+                        + str(scen_id))
+        
+class Car:
```
paasovaara: We could put these classes to separate .py files so they'd be more easy to find. Also in general case sometimes this can improve the decoupling of classes so that accidentally we don't access something that should be considered private.

Scripts/assignment.py
```diff
@@ -51,26 +51,22 @@ def assign(self, scen_id, time_period, matrices):
...
         self.logger.info("Assignment starts...")
         self.set_matrices(matrices)
-        function_file = os.path.join(self.path,
-                                     "d411_pituusriippuvaiset_HM30.in")
-        process = self.emme_modeller.tool(
-            "inro.emme.data.function.function_transaction")
-        process(function_file)
         self.calc_road_cost(scen_id)
         self.assign_cars(scen_id, param.stopping_criteria_coarse)
+        self.calc_boarding_penalties(scen_id)
```
paasovaara: We could consider if we want to use python convention of private methods with prefix self.__method_name() instead of self.method_name(). This could make it more intuitive which methods are considered to be part of the public interface of this class, == which methods should I use to process some data.

PR #9, opened 2019-04-05
Feat/pipenv

paasovaara:
This PR would bring in pipenv to help separate our libraries in local development machine (and possibly in the prod version also).


paasovaara:
@zptro it's also interesting that when importing OMX via emme, this is valid:

```   
import omx
```   

but it refuses to import when using official PyPi package. it's also in their official documentation. It

```   
import openmatrix as omx
```   

This probably means that EMME has included omx into their deployments before it has been released via PyPi.

PR #13, opened 2019-04-25
Feat/freight

zptro:


Scripts/demand/freight.py
```diff
@@ -55,9 +72,9 @@ def fratar(self, prod, trips, max_iter = 10):
...
 
     def calibrate(self, b, n, s):
-        """Calibrate a vector
-        b = true value for base
-        n = forecast for base
-        s = forecast to calibrate
+        """Calibrate a vector\n
+        b = true value for base\n
+        n = forecast for base\n
+        s = forecast to calibrate\n
```
paasovaara: Best practice: rename the variables to be more descriptive so you don't need comments here.

```   
def calibrate(self, true_value_for_base, forecast_for_base, forecast_to_calibrate):
```   

Number 5 is also a bit mystery here, it could be declared as a const: 
```   
SOME_HUMAN_READABLE_NAME=5
return numpy.where(s <  SOME_HUMAN_READABLE_NAME *n, s * b/n, s +  SOME_HUMAN_READABLE_NAME*(b - n))
```

Scripts/demand/freight.py
```diff
@@ -15,17 +15,34 @@ def calc_freight_traffic(self, mode):
...
+        # external_base = production_calib[len(production):]
+        # external_growth = 1.1
+        # external_production = external_growth * external_base
+        # production = numpy.append(production, external_production)
+        zone_numbers = self.base_demand.get_zone_numbers()
+        mtx = pandas.DataFrame(base_mtx, zone_numbers, zone_numbers)
+        prod = pandas.Series(production, zone_numbers)
+        # If forecast>5*base, destination choice is replaced by area average
+        cond = production_forecast < 5*production_base
+        for i in xrange(0, 30):
```
paasovaara: It's not obvious where does the numbers 30 and 5 come from, perhaps use constant names?

SOME_INDEX_TO_STOP=30

Scripts/demand/freight.py
```diff
@@ -15,17 +15,34 @@ def calc_freight_traffic(self, mode):
...
                                     production_base, 
                                     production_forecast)
-        # External demand is generated for the last centroids
-        external_base = production_calib[len(production):]
-        external_production = external_growth * external_base
-        production = numpy.append(production, external_production)
-        # Add harbour traffic!
-        demand = self.fratar(production, base_mtx)
+        # # External demand is generated for the last centroids
+        # external_base = production_calib[len(production):]
```
paasovaara: I would just remove commented code, we can always check it out of version control if needed.

Scripts/data_handling.py
```diff
@@ -5,16 +5,19 @@
...
     def __init__(self):
         # Here we should actually set up a link to data files
-        self.base_demand = numpy.arange(49).reshape(7, 7)
+        self.base_demand = numpy.arange(25).reshape(5, 5)
     
     def get_data(self, mode):
         return self.base_demand
 
+    def get_zone_numbers(self):
+        return [5, 6, 7, 2792, 16001]
```
paasovaara: Are these numbers some references to EMME constants or..? They could be declared as const variables with human readable names.

zptro: Added # TODO to make it clearer that this definition is just temporary

Scripts/demand/freight.py
zptro: Made the code more flexible, allowing arbitrary zone numbers.

PR #19, opened 2019-05-15
Changed first waiting time headway fraction to 0.3

zptro:


Scripts/parameters.py
```diff
@@ -332,12 +332,9 @@
     "penalty": "@wait_time_dev",
     "perception_factor": 3.5
 }
-waiting_time = {
-    "headway_fraction": 0.5,
-    "effective_headways": "hdw",
-    "spread_factor": 1,
-    "perception_factor": 1.5
-}
+first_headway_fraction = 0.3
```
paasovaara: Currently we use parameters.py as a place to declare global variables. there's always a risk that some part in the code accidentally modifies these variables and something breaks and the symptoms can be very strange. Also those variables can shadow local variables easily. 

We could consider having the parameters nested inside one (immutable) class / object which would then be given as constructor argument (or a global object, but it would be clearer that changing this variable is global, not just a local variable)..?

Issue #2, opened 2019-03-26
Wrap logging to some interface


paasovaara:
It would be useful to have a way to configure logging differently when running scripts via EMME (emme logging) and externally (F.ex during development, tests, etc).

We could use [Python Abstract base classes](https://docs.python.org/2/library/abc.html) for this purpose and have separate implementations for basic system.out-logging (perhaps to external file) and via EMME.

PR #26, opened 2019-05-23
Feat/application main

paasovaara:
- Create main structure for application-loop
- Reads configurations from json files, so that we can easily have multiple configurations and use f.ex app arguments to select which to use.
- Allows running in EMME-mode or without it
- wraps logging under one configurable logging interface. fixes https://github.com/HSLdevcom/helmet-model-system/issues/2

There's a lot of placeholders and lot of things to implement but this could act as the skeleton.


Scripts/helmet_app.py
```diff
@@ -0,0 +1,81 @@
...
+                    break
+
+        self.logger.info("All done, thank you!")
+
+    def handle_error(self, msg, exception):
+        self.logger.error(msg, exception)
+        fatal = True
+        return fatal
+
+    def estimate(self):
```
zptro: _Estimate_ is not the right word here. _Run_ or _simulate_ would be better.

PR #40, opened 2019-07-02
Feat/simple logit

johpiip:
Implementing a multinomial logit model for modeling origin in non-home-based tours. It is my first pull request to `helmet-model-system` so @zptro will hopefully make a sanity check? 😉


Scripts/demand/trips.py
```diff
@@ -11,6 +11,7 @@ def __init__(self, zone_data):
 
     def calc_demand(self, purpose, impedance):
         prob = self.calc_prob(purpose, impedance)
+        origin_prob = self.calc_origin_prob(purpose, impedance)
```
zptro: I suggest we create a new purpose, say "sop - source-other-periperal". Then we add an if-clause to self.calc_prob that redirects to the new function self.calc_origin_prob.

johpiip: https://github.com/HSLdevcom/helmet-model-system/pull/40/commits/f50d84ec85ae84a41d6a97ec1f1dee581a81acc4

johpiip:
Two issues left to @zptro:

1. What is returned when `calc_origin_prob()` is called with `oop` as purpose?
2. Probability to choose an origin should be mode-independent but tests require probability dictionaries names "transit".

Scripts/demand/trips.py
```diff
@@ -150,4 +153,39 @@ def calc_dest_util(self, purpose, mode, impedance):
...
+        b = parameters.origin_choice[model]["attraction"]
+        for i in b:
+            utility += b[i] * self.zone_data.values[i]
+        return utility
+
+    def calc_origin_prob(self, purpose, impedance):
+        utility = self.calc_origin_util(purpose, impedance)
+        exps = numpy.exp(utility)
+        # Here, size means kokotekija in Finnish
+        size = numpy.zeros_like(exps) + 1
```
zptro: ones_like()

johpiip: Thanks! https://github.com/HSLdevcom/helmet-model-system/pull/40/commits/e6aa5b2cc757a60a6793d2a346c08429a1a60678

PR #50, opened 2019-07-16
Feat/second destination

zptro:


Scripts/models/logit.py
```diff
@@ -0,0 +1,155 @@
...
+        b = self.dest_choice_param
+        utility = self.calc_origin_util(impedance)
+        exps = numpy.exp(utility)
+        # Here, size means kokotekija in Finnish
+        size = numpy.ones_like(exps)
+        size = self.add_zone_util(size, b["size"])
+        exps *= numpy.power(size, b["log"]["size"])
+        expsums = numpy.sum(exps, axis=0)
+        prob = {}
+        # Mode is needed here to get through tests even
```
ktuukkan: If the mode is indeed unnecessary, I'd fix the tests according to current implementation.

PR #59, opened 2019-09-17
Updated assignment_test so it works with new configuration

zptro:


TimoElolahde:
Running the updated assgnment_test from command prompt ended successfully without any problems. 
Using Helmet-UI gave first the same error message of missing emmebank as before. When it was corrected also Helmet-UI simulation was successful.

PR #71, opened 2019-10-15
Print results and matrices

zptro:
Also simplified handling of zone and matrix data
Split transit and bike matrices into work and leisure matrices to be able to do separate CBA calculations


johpiip:
@zptro The tests are failing because `transport_classes` does not exist in `parameters.py`.

zptro:
> 
> 
> @zptro The tests are failing because `transport_classes` does not exist in `parameters.py`.

Yes, commited wrong file. :(

Now fixed.

PR #85, opened 2019-11-11
modified generation code for script parameters.py

TimoElolahde:
modified generation code (variables and values) for script parameters.py


zptro:
In my opinion, best practice would be to add empty "zone" entries to all tour groups that do not have any zone variables. In this way the program does not just ignore all variables if "zone" is spelt wrong.

TimoElolahde:
If there are no individual dummys in the sub model (e.g. util3), should there be an empty individual_dummy entry as well?

zptro:
> 
> 
> If there are no individual dummys in the sub model (e.g. util3), should there be an empty individual_dummy entry as well?

Yes.

zptro:
Looks very good! Will merge this.

PR #86, opened 2019-11-14
Updated dependencies

zptro:
Updated pipenv dependencies to match Emme 4.4 python library versions. This removed the need for separate static dependencies, now when openmatrix is a stable version, so now all dependencies are installed via pipenv.

The new version of numpy might cause logit generation to crash, will see after merging.


zptro:
> 
> 
> Great! I updated pipenv on my computer and the tests passed. However, I also tried to run the model through Helmet UI but there I got `ImportError: No module named openmatrix`. Also, Emme version should be also updated to Helmet-UI config file: https://github.com/HSLdevcom/helmet-ui/blob/d4f4ce1cbb5fd20d9b1952db4df459e484e1cd5b/src/config.js

If you change python reference in UI Asetukset to the new Emme program folder, it should work again. When installing the UI from scratch, it looks for python in environment variable emmepath, so then it should suggest the correct python version.

I am not sure what the config file in UI actually uses that Emme version specification for.

johpiip:
Thanks, @zptro! I put a ticket to IT to upgrade my Emme, I forgot to do that earlier. :) So, for the time being, I can run tests but not the whole model system.

PR #88, opened 2019-11-18
modified coefficients of non-home based trips

TimoElolahde:


zptro:
Excellent! Now we need to copy the parameters to wo as well.

PR #93, opened 2019-11-25
Added call to last iteration

zptro:


zptro:
To be able to run this, the space for extra attributes needs to be expanded!

PR #100, opened 2019-11-29
Printing car users shares

annika-wsp:
I guess this could be merged now. Please comment on any style issues too for future reference.


zptro:
Looks very good!

PR #105, opened 2019-12-10
Add convergence checks for tour sums and car time impedance, printed …

annika-wsp:
… to txt files

### DIFF
--- a/Scripts/modelsystem.py
+++ b/Scripts/modelsystem.py
@@ -3,6 +3,7 @@ import datahandling.resultdata as result
 import assignment.departure_time as dt
 from datahandling.zonedata import ZoneData
 from datahandling.matrixdata import MatrixData
+import datahandling.resultdata as result
 from demand.freight import FreightModel
 from demand.trips import DemandModel
 from demand.external import ExternalModel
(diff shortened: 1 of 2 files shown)


annika-wsp:
Checks are failing, but I have no idea what is being checked...

Scripts/modelsystem.py
```diff
@@ -185,6 +195,24 @@ def run(self, impedance, is_last_iteration=False):
...
+        #  for calibration, and should be implemented differently later)
+        car_times = {}
+        car_times["aht_od_1"] = impedance["aht"]["time"]["car_work"][
+            self.zdata_base.zone_index(5),self.zdata_base.zone_index(7)]
+        car_times["aht_od_2"] = impedance["aht"]["time"]["car_work"][
+            self.zdata_base.zone_index(7),self.zdata_base.zone_index(5)]
+        car_times["iht_od_3"] = impedance["iht"]["time"]["car_work"][
+            self.zdata_base.zone_index(6),self.zdata_base.zone_index(7)]
+        car_times["iht_od_4"] = impedance["iht"]["time"]["car_work"][
+            self.zdata_base.zone_index(7),self.zdata_base.zone_index(6)]
```
zptro: You could make this a for loop, for instance looping over list of lists [["aht", 5, 7], ["aht", 7, 5]...]. The dict keys in car_times can be concatenated from these lists.

annika-wsp: That can be done quite easily I guess, but the problem is, where should the list be saved? And should we have some kind of "switch" in the config file to choose whether the test network or the real network is used?

zptro: I think we could put the list in a file in the folder Scenario_input_data. Then no switch is needed, only zones that exist in the current scenario will be evaluated.

annika-wsp: Ok, I'll try to do that, and we'll see how it goes...

PR #110, opened 2019-12-18
Update generation constants

johpiip:
- Add modes to `hoo` generation
- Update generation constants to fix an error with weights
- Calculate `oo` generation in `helmet-data-preprocessing`


johpiip:
@zptro Check out `oo` parameters: they are very different from your values. I don't know the reason, maybe there was a mix-up before? The largest difference is shopping-based tours: to me, it sounds reasonable that people do not make non-home-based tours which would start from shopping locations.

PR #111, opened 2019-12-19
Implementation of method of successive averages for impedance matrices

zptro:
This can be used for averaging impedance matrices


atte-wsp:
Replaced with new PR.

PR #113, opened 2020-01-03
Add convergence checks for tour sums and car time impedance, second t…

annika-wsp:
…ry. Results are printed into txt files. There we some changes to modelsystem.py in between, so especially the tours sums part should be checked. Od pairs are in a separate file under Scenario_input_data.


annika-wsp:
If I interpret the test log correctly, the tests fail when testing agent_model. That would also explain why the changes work just fine when I use the code myself. In practice it means that I have no idea how to fix this, since I have no idea how the agent model works or is different from the aggregated model -> I'll just leave it like this for now and the code can be found here later when the convergence is revisited.

atte-wsp:
No longer needed as focus of convergence test changed.

PR #117, opened 2020-01-14
Restructured printing so that aggregated matrices are for tours by car

zptro:
Restructured printing so that aggregated matrices are for tours by car, including car passengers


zptro:
> 
> 
> Otherwise fine by me, but a question about the trip lengths. Is this (the last change in intervals) the correct place to change the trip length intervals for calibrations, or is there another place somewhere? Does this affect something other than the printed intervals txt file? Johanna's data for calibration now has ("0-1", "1-3", "3-5", "5-10", "10-20", "20-30", "30-40", "40-inf"), so I had tried to change the intervals here.

Thanks for noticing! I had used an old version of the file by accident.

PR #120, opened 2020-01-20
Feat/car density model

johpiip:
I will implement the car density model here (car ownership, cars per 1000 people). Not yet ready, but we can discuss about the execution here.


johpiip:
Not ready to be merged. Still some problems:

- controlling which zones get new predicted car densities and which still use the input data densities
- getting a base density from the demand in base matrices

johpiip:
Other problems solved but these are still pending:

- Car density model does not work on test network because there probably is not any transit costs (ticket zones) associated with the zones.
- On the first iteration, car model is not used but `.car` data from `Scenario_input_data` is used instead. The car model is used on following iterations. Can `time_ratio` and `cost_ratio` be deduced from the base data so that the model could be ran in first iteration too?

zptro:
>     * On the first iteration, car model is not used but `.car` data from `Scenario_input_data` is used instead. The car model is used on following iterations. Can `time_ratio` and `cost_ratio` be deduced from the base data so that the model could be ran in first iteration too?

This should be easy to fix. Base impedances (based on base demand) are calculated in `assign_base_demand`. You could just put a call to `_update_ratios` there!

johpiip:
Thank you! The car density model now works for all residents and zones inside metropolitan area. The tests fail because the cost variable is 0 everywhere.

zptro:
Direct model finished

Scripts/modelsystem.py
```diff
@@ -90,9 +93,15 @@ def assign_base_demand(self, use_fixed_transit_cost=False):
         for tp in parameters.emme_scenario:
             self.logger.info("Assigning period " + tp)
             with self.basematrices.open("demand", tp) as mtx:
-                base_demand = {ass_class: mtx[ass_class] for ass_class in self.ass_classes}
-            self.ass_model.assign(tp, base_demand)
+                for ass_class in self.ass_classes:
+                    self.dtm.demand[tp][ass_class] = mtx[ass_class]
+            # Base matrices do not provide "transit_work" as its own assignment
+            # class but "transit" is close enough.
+            self.dtm.demand[tp]["transit_work"] = self.dtm.demand[tp]["transit"]
```
zptro: Need to unify this fix with Atte's fix?

Scripts/models/linear.py
```diff
@@ -0,0 +1,105 @@
...
+                    i, self.bounds, generation, zdata.SURROUNDING_AREA)
+                prediction[:k] += b[i][0] * data_capital_region
+                prediction[k:] += b[i][1] * data_surrounding
+        return prediction
+
+    def _add_log_zone_terms(self, prediction, b, generation=False):
+        zdata = self.zone_data
+        for i in b:
+            prediction += b[i] * numpy.log(zdata.get_data(i, self.bounds,
+                generation))
```
atte-wsp, line 52: Share of detached houses from self.zonedata include zeros (full model), so this would calculate log(0), which causes a warning. Could there be clip in zonedata.py?

Scripts/parameters.py
```diff
@@ -2879,6 +2879,18 @@
         ("age_65-99", "male"): 1.8737,
     },
 }
+car_density = {
+    "constant": 0.0,
+    "generation": {
+        "share_detached_houses": 2.408e-01, # Originally estimated for percentages (0-100), here transformed
```
atte-wsp, line 2885: Zone data input values for "share_detached_houses" are in range 0-100 (scenario input data in Sharepoint). Is this correct or should the values be between 0-1? I think this might cause the overflow in demand matrices.

Scripts/models/linear.py
```diff
@@ -0,0 +1,105 @@
...
+        
+        # print car density by municipality
+        prediction_municipality = []
+        for municipality in parameters.municipality:
+            i = slice(parameters.municipality[municipality][0],
+                      parameters.municipality[municipality][1])
+            # comparison data has car user shares of population
+            # over 6 years old (from HEHA)
+            prediction_municipality.append( car_density.loc[i].sum() 
+                                    / population_7_99.loc[i].sum())
```
atte-wsp, line 89: Sum of results are not reasonable. Should the numerator be product of car density and population by zone (weighted average)?

Scripts/models/linear.py
```diff
@@ -0,0 +1,105 @@
...
+                          
+        # print car density by area (to get Helsinki CBD vs. Helsinki other)
+        prediction_area = []
+        for area in parameters.areas:
+            i = slice(parameters.areas[area][0],
+                      parameters.areas[area][1])
+            # comparison data has car user shares of population
+            # over 6 years old (from HEHA)
+            prediction_area.append( car_density.loc[i].sum()
+                            / population_7_99.loc[i].sum())
```
atte-wsp, line 102: Sum of results are not reasonable. Should the numerator be product of car density and population by zone (weighted average)?

Scripts/parameters.py
zptro: Yes, I think the values should be 0-1

Scripts/models/linear.py
zptro: But detached houses variable is not log (?)

zptro: But detached houses variable is not log (?)

Scripts/parameters.py
zptro: Yes, I think the values should be 0-1

Scripts/models/linear.py
zptro: @johpiip

johpiip: Does this work? https://github.com/HSLdevcom/helmet-model-system/pull/120/commits/ad85e88d710e3cc92312236adfd5dc326e8e13c1

Scripts/models/linear.py
johpiip: Does this work? https://github.com/HSLdevcom/helmet-model-system/pull/120/commits/ad85e88d710e3cc92312236adfd5dc326e8e13c1

johpiip:
> 
> 
> Full model system run ends up with overflow error, as the model tries to import infinite / NaN demand matrices to Emme-project. I think the reason is in the input values of share of detached houses. More detailed comments below.

Was this already solved, @atte-wsp ? :)

johpiip:
Hey, should the weighted mean I fixed in https://github.com/HSLdevcom/helmet-model-system/pull/120/commits/ad85e88d710e3cc92312236adfd5dc326e8e13c1 be actually for whole population (0-200 years old)? I think car density is defined as "the number of cars per inhabitant" and inhabitants include *all* people, not just people over 6 years old. Isn't the input data also defined in the same way?

There is also a variable "cars_per_1000_people" which is just 1000 times whatever this model predicts.

Scripts/models/linear.py
atte-wsp: Yes, it's working now! :)

atte-wsp:
> > Full model system run ends up with overflow error, as the model tries to import infinite / NaN demand matrices to Emme-project. I think the reason is in the input values of share of detached houses. More detailed comments below.
> 
> Was this already solved, @atte-wsp ? :)

Yeah this was solved, changing the detached houses parameter (0-1) fixed the problem.

atte-wsp:
Model works without errors now. I tried the model run with three iteration rounds and resulting car use shares were quite high compared to data. This might be because the transit / car impedance hasn't yet converged so I will check these again with 10 iteration rounds.

PR #123, opened 2020-01-22
Fix/sop

johpiip:
This PR fixes the calculation of probabilities in OriginModel ("sop"). Parameters have been updated. Parameter values where found and organized by @TimoElolahde, thanks! The new script handles mode-dependent exponent terms (`exponent` in `parameters.py`) in a new method, `_add_exponent_terms`.

Changes to `OriginModel` code have been quite large, especially regarding exponent terms, so I am waiting for @zptro's review before merging to master.

### DIFF
--- a/Scripts/models/logit.py
+++ b/Scripts/models/logit.py
@@ -365,12 +365,7 @@ class OriginModel(LogitModel):
                 Choice probabilities
         """
         b = self.dest_choice_param
-        utility = self._calc_origin_util(impedance)
-        exps = numpy.exp(utility)
-        # Size = kokotekija
-        size = numpy.ones_like(exps)
-        size = self._add_zone_util(size, b["size"])
-        exps *= numpy.power(size, b["log"]["size"])
+        exps = self._calc_exps(impedance, b)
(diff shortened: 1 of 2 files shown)


zptro:
This is probably a working solution!

However, I actually like Timo's solution better, as it does not require any new code. Virpi's "new" model is a [normal nested logit model](https://github.com/HSLdevcom/helmet-model-system/blob/80f92d8dd2e1f763dee8a1cbe4f08bfb05de64ad/Scripts/models/logit.py#L300). Sorry for not noticing this change, which occurred when I was away in August.

johpiip:
@zptro: yes, Timo's solution organizes the variables clearer but it does not pass the tests as it is. I need to take a new look at it.

zptro:
> 
> 
> @zptro: yes, Timo's solution organizes the variables clearer but it does not pass the tests as it is. I need to take a new look at it.

Correct. The used class OriginModel would need to be changed to ModeDestModel, and possibly something else too.

johpiip:
Continuing the work in a new branch: https://github.com/HSLdevcom/helmet-model-system/tree/feat/sop2

PR #127, opened 2020-01-29
Feat/sop2

johpiip:
This is a new version of the `sop` model which chooses the origin zone for non-home-based tours of the people living in peripheral municipalities. This replaces and closes #123 with another solution.

Otherwise this works, but some help is needed:

- This PR introduces modes car and transit to sop model. Previously, the probability dictionary was a dictionary with only one mode, `all`. Does this affect the origin choice?
- If the modes are now "real" modes, does this affect the assignment? Virtual tours generated in `sop` model should not be assigned.

### DIFF
--- a/Scripts/models/logit.py
+++ b/Scripts/models/logit.py
@@ -355,7 +355,7 @@ class SecDestModel(LogitModel):
         return prob
 
 
-class OriginModel(LogitModel):
+class OriginModel(DestModeModel):
     def calc_prob(self, impedance):
         """Calculate matrix of choice probabilities.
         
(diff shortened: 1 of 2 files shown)


johpiip:
And one more thought:

- Does having real mode names affect the calibration? If the calibration Excels consider mode "all" to be discarded but not normal modes?

zptro:
> 
>     * This PR introduces modes car and transit to sop model. Previously, the probability dictionary was a dictionary with only one mode, `all`. Does this affect the origin choice?
> 
>     * If the modes are now "real" modes, does this affect the assignment? Virtual tours generated in `sop` model should not be assigned.

This should be fine! If <purpose.dest == "source">, the demand is not assigned.

PR #130, opened 2020-02-10
Feat/calibration

annika-wsp:
Merge calibrated parameters for generation, destination choice and mode choice into master (first version, some of these might have to be calibrated further later).


atte-wsp:
Calibration ok based on testing and traffic volumes comparison. New calibration might be needed at some point.

PR #132, opened 2020-02-11
Update parameters.py

zptro:
Removed hard-coded emme scenario numbers from parameters.py


Mew-www:
Merging since zptro is currently unavailable (for a week).

PR #135, opened 2020-02-19
Include T_radius to sop model

johpiip:
Based on email on 18th of February at 13:08.


johpiip:
@atte-wsp @zptro I will merge this because this fixes a bug in `master`.

PR #147, opened 2020-03-15
Revert logit

atte-wsp:
Undo: February 7 "Avoid side effects when calculating parts of utility". This merge causes the full model system to produce erroneous assignment results. I haven't yet figured out the reason, but would merge this to the master, as I think it should be functional for testing purposes.

### DIFF
--- a/Scripts/models/logit.py
+++ b/Scripts/models/logit.py
@@ -68,7 +68,7 @@ class LogitModel:
         size = numpy.zeros_like(utility)
         size += self._add_sec_zone_util(size, b["size"])
         impedance["size"] = size
-        self._add_sec_zone_util(size, b["size"])
+        self._add_log_impedance(dest_exps, impedance, b["log"])
         if mode != "logsum":
             threshold = parameters.distance_boundary[mode]
             dest_exps[impedance["dist"] > threshold] = 0
--- a/Scripts/models/logit.py
+++ b/Scripts/models/logit.py
@@ -24,13 +24,13 @@ class LogitModel:
         for mode in self.mode_choice_param:
             b = self.mode_choice_param[mode]
             utility = numpy.zeros_like(expsum)
-            utility += self._add_constant(utility, b["constant"])
-            utility += self._add_zone_util(
+            self._add_constant(utility, b["constant"])
+            utility = self._add_zone_util(
                 utility.T, b["generation"], generation=True).T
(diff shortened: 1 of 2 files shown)


zptro:
@atte-wsp, could you fix this and merge asap?

Scripts/models/logit.py
```diff
@@ -62,19 +62,19 @@ def _calc_dest_util(self, mode, impedance):
...
-        utility += self._add_sec_zone_util(utility, b["attraction"], orig, dest)
-        utility += self._add_impedance(utility, impedance, b["impedance"])
+        self._add_sec_zone_util(utility, b["attraction"], orig, dest)
+        self._add_impedance(utility, impedance, b["impedance"])
         dest_exps = numpy.exp(utility)
         size = numpy.zeros_like(utility)
         size += self._add_sec_zone_util(size, b["size"])
         impedance["size"] = size
-        dest_exps *= self._add_log_impedance(dest_exps, impedance, b["log"])
+        self._add_sec_zone_util(size, b["size"])
```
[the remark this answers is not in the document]
atte-wsp: That's correct! There was simple copy / paste mistake. I should have been more careful before commit. Now corrected.

Scripts/models/logit.py
```diff
@@ -62,19 +62,19 @@ def _calc_dest_util(self, mode, impedance):
     def _calc_sec_dest_util(self, mode, impedance, orig, dest):
         b = self.dest_choice_param[mode]
         utility = numpy.zeros_like(next(iter(impedance.values())), self.dtype)
-        utility += self._add_sec_zone_util(utility, b["attraction"], orig, dest)
-        utility += self._add_impedance(utility, impedance, b["impedance"])
+        self._add_sec_zone_util(utility, b["attraction"], orig, dest)
+        self._add_impedance(utility, impedance, b["impedance"])
         dest_exps = numpy.exp(utility)
         size = numpy.zeros_like(utility)
         size += self._add_sec_zone_util(size, b["size"])
```
zptro: This is still done in Johanna's style?

zptro: This is still done in Johanna's style though?

atte-wsp: Checked this one from commits and that's correct. I already merged this pull request, so it seems that I need to do another pull request to fix this.

PR #149, opened 2020-03-16
Implementation of car density change model

zptro:
Car ownership model is applied only for population growth. Exising population keeps its current car density.


atte-wsp:
Found interesting error here. Python 2.7 performs automatically rounding to nearest integer when diving integer values (integer division). Population datatype is int, so calculating growth shares of population on car density change model result only zeros.

This might affect also other model parts, so I think this should be solved by converting Zonedata values to floats. Also error checks of input data could be made at the same time. I'll make separate PR for Zonedata datatypes.

TimoElolahde:
Good point! In the beginning of this year I fixed some calculation formulas of e.g. transit cost (x/44 replaced by x/44.0), but nobody noticed the effect of datatypes

PR #154, opened 2020-04-14
Refactor emme assignment

atte-wsp:


Scripts/assignment/emme_assignment.py
```diff
@@ -448,7 +438,7 @@ def _specify(self):
                 van.spec,
             ],
             "background_traffic": {
-                "link_component": param.background_traffic,
+                "link_component": "ul3",
```
zptro, line 441: Why?

Scripts/assignment/emme_assignment.py
```diff
@@ -747,7 +736,9 @@ def _assign_transit(self, scen_id, count_zone_boardings=False):
         emmebank = self.emme_project.modeller.emmebank
         scenario = emmebank.scenario(scen_id)
         self.emme_project.logger.info("Transit assignment started")
-        self.emme_project.transit_assignment(self.transit_spec, scenario)
+        self.emme_project.transit_assignment(
+            specification = self.transit_spec,
```
zptro, line 740: Remove spaces.

Scripts/assignment/emme_assignment.py
```diff
@@ -763,6 +754,26 @@ def _assign_congested_transit(self, scen_id):
...
+            "result": mtx_path,
+            "expression": "999999*(p!=q)",
+            "constraint": {
+                "by_value": {
+                    "od_values": mtx_time,
+                    "interval_min": 0,
+                    "interval_max": 999999,
+                    "condition": "EXCLUDE"}
+                    }
+                })
```
zptro, line 779: I have tried to avoid doing matrix operations inside emme, for efficiency and readability reasons. I think this could be done according to the same logic as for other matrix modifications done in `get_impedance()`.

Scripts/parameters.py
```diff
@@ -168,15 +168,33 @@
     'a',
     's',
 ]
+# Link attributes initialized 
+# in helmet-model-system 
+emme_attributes = {
+    "@bus": "LINK",
+    "@rumsi": "LINK",
+    "@ruma": "LINK",
```
zptro, line 176: Change these as well to self-explanatory names.

zptro:
Travis notfications have for some reason not been working in the last few weeks, but tests for this branch do crash. So need to check how well MockAssignment works after these changes.

Scripts/assignment/emme_assignment.py
atte-wsp: For some reason when accessing link parameters from Network API (as in module _calc_background_traffic) the name of this parameter is "data3" and in assignment the same parameter is called "ul3". I could not find any smart way around this issue when using imported parameters, so decided to hard code "ul3/data3" to Emme-assignment. Are there some drawbacks that I didn't think about?

Scripts/assignment/emme_assignment.py
atte-wsp: Of course! I just wonder why I didn't think of this at first place! :) The operation is much more simple and compact with Numpy.

Scripts/assignment/emme_assignment.py
zptro: Maybe it could be called ul3 in parameters, but we would replace "ul" with "data" where necessary? I think it would be nice if this would not be hardcoded.

Scripts/assignment/emme_assignment.py
```diff
@@ -433,11 +404,30 @@ def calc_transit_cost(self, fares, peripheral_cost, default_cost=None):
...
+            value_of_time_inv = 0.2, 
+            link_costs = "length",
+            demand_mtx = self.demand_mtx, 
+            result_mtx = self.result_mtx)
+        trailer_truck = Car(
+            ass_class = "trailer_truck", 
+            value_of_time_inv = 0.2, 
+            link_costs = "length",
+            demand_mtx = self.demand_mtx, 
+            result_mtx = self.result_mtx)
```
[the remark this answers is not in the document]
zptro: Still using spaces though...

PR #155, opened 2020-04-14
Enable Github Actions

zptro:
@HSLdevcom/transport-modelling has obviously overused its free account on Travis. But we can use Github Actions for running the tests instead.


yuruki (approved):
Yes, actually all public repo travis-ci pipelines are encouraged to migrate to GitHub Actions. 👍

TimoElolahde:
OK

PR #156, opened 2020-04-16
Feat/update readme

johpiip:
I rewrote the whole README after my experiences. The order is now 1) production instructions, and 2) development instructions. A couple of questions:

1. Is [OMX-README.md](https://github.com/HSLdevcom/helmet-model-system/blob/master/OMX-README.md) still relevant or should I incorporate it to README?
2. Are following files needed:
    - [import-dev-dependencies.bat](https://github.com/HSLdevcom/helmet-model-system/blob/master/Scripts/import-dev-dependencies.bat)
    - [import-python.bat](https://github.com/HSLdevcom/helmet-model-system/blob/master/Scripts/import-python.bat)
    - [run-emme-python.bat](https://github.com/HSLdevcom/helmet-model-system/blob/master/Scripts/run-emme-python.bat)
    - [run-pipenv.bat](https://github.com/HSLdevcom/helmet-model-system/blob/master/Scripts/run-pipenv.bat)
    - [run-python.bat](https://github.com/HSLdevcom/helmet-model-system/blob/master/Scripts/run-python.bat)
    - [run-tests.bat](https://github.com/HSLdevcom/helmet-model-system/blob/master/Scripts/run-tests.bat)
3. Was there a test script (other than tests) which would work without an Emme license? At least `test_assignment.py` nor `helmet.py` are not working even after I changed `"USE_EMME": false`.

### DIFF
--- a/OMX-README.md
+++ b/OMX-README.md
@@ -1,8 +1,0 @@
-# OMX 
-
-Emme supports OpenMatrix library for exporting matrices. 
-
-- More info here: https://github.com/osPlanning/omx/wiki/EMME 
-- Python source codes: https://github.com/osPlanning/omx-python
-- Useful tool for exporting from cmd line: https://github.com/bstabler/EMXtoOMX
-
--- a/README.md
+++ b/README.md
@@ -48,71 +48,48 @@ python test_assignment.py
 
 ## Development
 
-### Environment
+If you do not have INRO Emme license or you wish to develop HELMET 4.0 source code, you need to set up a development enviroment. Although not covered here, installing [Git](https://git-scm.com/downloads) is highly recommended!
 
-We have two execution environments:
-- deployed "production" environment
-- local development environment
+### Environment and dependencies
 
-In both cases we're using Python version 2.7 because our final deployment target (EMME) supports only 2.7.
+We are using Python 2.7 because it is supported by INRO Emme software.
 
-### Dependencies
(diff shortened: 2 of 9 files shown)


zptro:
1. OMX-README might not be very relevant, you can incorporate it.
2. Probably not very relevant.
3. Indeed, there were some inconsistencies when running helmet.py with mock assignment. I fixed these here: https://github.com/HSLdevcom/helmet-model-system/pull/151/commits/3e5fe0205a57ea97b475c6c8f0c0746e20c72302. Make sure you have matrix results in your chosen results folder (RESULTS_PATH/SCENARIO_NAME) for the same zones as you have zone data, this where MockAssignment gets its matrices.

johpiip:
Thanks, @zptro! For the mock assignment to work, two things need to be fixed. First, in ´helmet.py´, we can not import `EmmeProject` straight but we have to something like:

```
try:
    from emme_bindings.emme_project import EmmeProject
except:
    pass
```

Then, I noticed that he mock assignment handles ´.omx´ files differently to the Emme assignment. This has something to do with `trailer_truck` not being in `time_iht.omx`. Test `omx` files that come with the repo work fine, but not the ones that are in Sharepoint.

This is too wide a subject to fix in this PR so I'm going to merge this now. :) I added a README to Scripts folder too, and gathered stuff about dev-config.json file.

PR #160, opened 2020-04-29
Fix vehicle_kms calculation

samakinen:
A for loop in print_vehicle_kms() can skip initializing 'mode' variable and report the statistics for a wrong vehicle type.


Scripts/assignment/emme_assignment.py
```diff
@@ -305,6 +305,7 @@ def print_vehicle_kms(self, resultdata):
                         car_vol -= link[param.link_volumes[ass_class]]
                     kms["car"][vdf] += (param.volume_factors["car"][tp] * car_vol * link.length)
             for line in network.transit_lines():
+                mode = "other"
```
atte-wsp, line 308: Adding mode "other" causes error at line 313, because dictionary `transit_dists` does not have previous value set for this mode. I think operator `+=` requires value to be initialized before assignment.

samakinen: True, I had missed that when staging the changes. I created a new PR to fix it. I think the correct solution would be to use defaultdict here.

PR #165, opened 2020-05-11
Refactor Emme-assignment

atte-wsp:
* New classes that hold Emme-scenario specific methods. This makes it easier to assign single time periods, to perform extended analyses and use separate transit classes.
* No modifications needed to modelsystem. Only Emme-assignment is modified.
* Added 24h volume expansion to Emme-assignment (last iteration).

This update still needs some work, but I think this would be the right time to see if the changes are to right direction. I did a test run for refactored model and it produces the same results as master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,9 +47,10 @@ Database/aja_*.mac
 Database/d2*.in
 Logbook/project.mlbk-shm
 Logbook/project.mlbk-wal
-Matrices/2016/*
-Matrices/2030/*
-Matrices/base/*
+Network_builds/*
+Worksheets/*
+Specifications/*
+Views/*
 Scripts/.pytest_cache
 Scripts/tests/.pytest_cache
 Scripts/.env
--- a/Database/4verkkoa_HM40.mac
+++ b/Database/4verkkoa_HM40.mac
@@ -11,28 +11,29 @@
 ~#     - kaantymiset
 ~#     - seka linkkien ruuhkamaksut (e/km)
 ~#
-~#  parametrit p1 = vaihtoehdon tunnus (esim. 2030_20170329)
-~#             p2 = alikansio, josta tiedostot luetaan (nykyinen, jos puuttuu)
-~#             p3 pyöräverkon skenaario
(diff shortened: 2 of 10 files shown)


TimoElolahde:
There already are some time period depend attributes in emmebanks, e.g. @hinah|@hinih|@hinpt and @hwaht|@hwiht|@hwpt. In addition to them, also values of following attributes depend on time period.

Results of assignment, cannot be modified in network calculator:
Inode: fialii inboai 
Link: timau volad volau volax  
Turn: pvolau pvolad ptimau 
Segment: timtr

The values of these time period depend attributes can be modified in network calculator:
Link: ul1 ul2 ul3 vdf @ruma @rumpi @rumsi @kev @yhd @ka @pa @vm @kavl
Transit line: lay1 lay2 @vm1 @ajoai
Segment: ttf ttft us1 us2 voltr @wait_time_dev @ccost @base_timtr

If we want to save the detailed results of every time specific assignment in one scenario only, we need a lot of more extra attributes to save the values of all these input and output values. Otherwise some intermediate or detailed results are lost.

In addition to that there are several possibilities to report and analyze results of assignments in Emme (e.g. differences of volumes in two scenarios) but only if they are saved in default attributes. Of course, corresponding analyses can also be made using extra attributes, but I suppose it is more complicated.  If there are odd results (or doubts of problems) in assignment one can analyze the scenario itself where the calculations are made in stead of copies of attribute values. Maybe some intermediate or detailed results are not even copied into extra attributes.

atte-wsp:
Could this structure of dividing Emme-scenarios to separate classes be used even if we want to remove the time periods later? We probably will have bike and car/transit scenarios also in this case. I think that currently there is too much code in emme_assignment.py, which makes using it a bit difficult.

In a long run I think that best solution would be to keep separate Emme-scenarios for each time period. In our projects the extended analyses are used very often and having to create separate methods for using them would consume a lot of time and effort. However, I would prefer integrating the macro-network-methods to Python-Emme-assignment, so that user modifies only one network file, which is used to create others automatically.

PR #167, opened 2020-05-12
Distance cost from Zonedata

atte-wsp:
* Add car distance cost from zonedata
* Creating EmmeAssignment instance needs to be adjusted. Zonedata is read from modelsystem.py only after EmmeAssignment is initialized and assignment instance variable (number of zones) is used when reading zone data.
 * Changed Emme-network attributes and background variables to be prepared only when base assignment is performed. Using also is_first_iteration variable as a substitute for has_assigned_bike_and_walk as it is more understandable way of handling same issue.
* Fix Zonedata as the self.car_dist_cost was returning header instead of value.

Could @TimoElolahde check this Pull Request as I think Jens is unavailable. I did test run already with these changes and everything seems to be working fine.

### DIFF
--- a/Scripts/assignment/emme_assignment.py
+++ b/Scripts/assignment/emme_assignment.py
@@ -9,8 +9,7 @@ from datatypes.path_analysis import PathAnalysis
 
 
 class EmmeAssignmentModel(AssignmentModel, ImpedanceSource):
-    def __init__(self, emme_context, first_scenario_id, car_dist_cost=param.dist_unit_cost,
-                demand_mtx=param.emme_demand_mtx, result_mtx=param.emme_result_mtx):
+    def __init__(self, emme_context, first_scenario_id, demand_mtx=param.emme_demand_mtx, result_mtx=param.emme_result_mtx):
         """
         first_scenario_id (bike scenario) is usually #19,
             followed by (#20) walk scenario, (#21) morning scenario, (#22) midday scenario, and (#23) evening scenario.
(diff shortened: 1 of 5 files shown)


Scripts/assignment/emme_assignment.py
```diff
@@ -69,14 +71,17 @@ def assign(self, time_period, matrices, is_last_iteration=False, is_first_iterat
...
         self.set_emmebank_matrices(matrices)
         scen_id = self.emme_scenarios[time_period]
-        if not self._has_assigned_bike_and_walk:
+        if is_first_iteration:
+            self._prepare_network()
             self._assign_pedestrians(scen_id)
             self._assign_bikes(self.bike_scenario,
                             self.result_mtx["dist"]["bike"]["id"],
                             "all",
                             "@bike_"+time_period)
```
zptro, line 80: So we will run pedestrian and bike assignment three times, even if the results are the same every time. If this is acceptable when it comes to run time, then ok.

Scripts/datahandling/zonedata.py
```diff
@@ -38,8 +38,8 @@ def __init__(self, data_dir, zone_numbers):
         transit_zone["dist_fare"] = transit_zone["fare"].pop("dist")
         transit_zone["start_fare"] = transit_zone["fare"].pop("start")
         self.transit_zone = transit_zone
-        car_cost = read_csv_file(data_dir, ".cco", squeeze=True)
-        self.car_dist_cost = car_cost[0]
+        car_cost = read_csv_file(data_dir, ".cco", squeeze=False)
+        self.car_dist_cost = car_cost["dist_cost"][0]
```
zptro, line 42: This one worked before too, but ok then.

atte-wsp: The problem was that cco-files in test-data folder did not have headers and full model files in Sharepoint had. The files with header were read as an array with two values and index [0] returned the header string.

I thought that this would give more explicit error when reading zonedata, if header is not found. Otherwise the error would point to assignment, where model tries to multiply float with string.

Scripts/assignment/emme_assignment.py
atte-wsp: Something I didn't think about.. Good that you pointed it out! Checked from the log files that pedestrian and bike assignment take less than 10 seconds to complete.

I'll keep this change for now and take this into account when trying to optimize the run time as a whole.

PR #171, opened 2020-05-15
Fix zonedata value types

atte-wsp:
* Python 2.7 performs automatically rounding to nearest integer when diving integer values (integer division). This causes calculating growth shares of population on car density change model to result only zeros.
* This fix converts all numeric Zonedata values to floats to prevent latent integer divisions in model-system.
* Simultaneously check Zonedata values and raise ValueErrors.


atte-wsp:
> In addition to the modifications you made in the python code should the zone data input files be modified as well so that there is at least one decimal value in every column?

I think this is not necessary. This update should take care that model uses correct format for input data even if data is given as integers. Strings and other values cause an exception to be raised.

zptro:
The for-loop does not work because `frame` does not point back to the original DataFrame.

PR #172, opened 2020-05-18
Remove fixed scenario ids

atte-wsp:
* Currently method print_vehicle_kms uses fixed scenario ids (21-23). However these should be set by user via selecting first_scenario_id
* This fix removes emme_scenario variable from parameters as this should not be used as fixed variable in any case
* Emme-scenarios are set in Emme-assignment. Classes modelsystem and departure_time are set to use this instance variable

Fyi @samakinen if you're currently working with CBA


TimoElolahde:
So elegant!

PR #175, opened 2020-06-02
Fix/aggregation of car results

johpiip:
In this PR, the aggregation weights of car density are changed from 6-99-year-olds to the whole population. I think the 6-99-year-olds were there previously because they are used in car user shares (the amount of car drivers per 6-99-year-old population, HAP/EHAP model).


johpiip:
Can @atte-wsp check at some point of the simulations that the results of car densities are still on the same level? The test data is too sparse to make aggregation tests... The car density on different zones should not change because of this PR, but the aggregation to municipalities and areas can change.

PR #177, opened 2020-06-09
MSA method for car impedance

atte-wsp:
* Implementation of method of successive averages for car impedance matrices.
* This similar solution that was implemented before, but there was many updates to model system in between and it seemed easier to merge as new branch.
* I'll close the original PR and branch.

### DIFF
--- a/Scripts/modelsystem.py
+++ b/Scripts/modelsystem.py
@@ -110,6 +110,8 @@ class ModelSystem:
             impedance[tp] = self.ass_model.get_impedance()
             if tp == "aht":
                 self._update_ratios(impedance, tp)
+        
+        self.imptrans.average_car_impedance(impedance)
         return impedance
 
     def run_iteration(self, previous_iter_impedance, is_last_iteration=False):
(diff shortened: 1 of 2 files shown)


atte-wsp:
After convergence and CBA testing the MSA method seems unnecessary for model convergence. However, it could be used to optimize running time, if it is considered to be a problem. I'll mark this as draft.
* Trip matrices are converged nicely both with and without MSA. Both methods have difference of 0-5 trips in sums of mode, purpose and area specific matrices, when comparing iteration rounds 9 and 10.
* CBA calculation has total difference of 2-3 million euros between 9 to 10 rounds in both with and without MSA. However, these differences are related to stopping criteria of car assignment. The main issue is change of transit operating cost (vehicle time), which is not directly depended on method of calculating impedance matrices.
* Both methods converge to approximately same demand. Difference in mode specific demand is max 200 trips per day in whole model area.

Test runs were performed with MAL2019 year 2030 network and zone data.

TimoElolahde (dismissed):
Good news

PR #181, opened 2020-07-01
Expand peak hours to day volumes

atte-wsp:
* Add expand methods to Emme Assignment. The 24h expansions is calculated during last iteration
* Save volumes for different car classes, transit volumes and boardings and bikes
* Transit results use modeller tool Network results, so this is added to Emme bindings
* Add required extra attributes to parameters.py


atte-wsp:
There seems to be a problem with Travis tests, which are not working. This can be merged when the tests are running again.

PR #182, opened 2020-07-02
Test result methods

atte-wsp:
This PR contains methods that have been used for testing and calculating results for report. They are implemented in separate `helmet_test_methods.py` main function, which uses the same config files that are used to run `helmet.py`.

Methods:
* Matrix aggregation (municipalities and postal codes)
* End assignment for Helmet-scenario (with previously calculated result matrices). Enables extended analyses and does not overwrite matrices in emmebank
* 24h expansion for result matrices
* Count data import 
* Transit strategy analysis for trips with transfers and specific mode

I think some of these methods could be useful for other user as well and they could be merged to distributed model system. However, interface is not yet user friendly and not all methods are needed widely, so additional work is still needed.


atte-wsp:
I'll close this one as well as most methods would need to be updated.

PR #183, opened 2020-07-02
Transit work leisure

atte-wsp:
Enabling separate assignment of transit work and leisure. The assignments are performed in sequence so this probably increases the model running time. However, separating work / leisure could be beneficial to improve two issues found in model testing:

1) Expansion of 24h transit volumes underestimates the volumes of trunk network (frequent service during evenings and early morning). This model has different coefficients for work and leisure that might help this issue, because work and leisure trips have different spatial distribution.
2) Transfer penalties and number of transfers per trip are overestimated. This leads to number of boardings to be clearly overestimated. This change could help because work trips have almost double the amount of transfers when comparing to leisure trips (HSL 2015: Selvitys joukkoliikenteen vaihtovastuksesta) and the change enables to use different penalties for different trip purposes. 

These however have not yet been tested. Also coding requires still work as there might be more elegant solutions to specify transit assignment.


atte-wsp:
Update: I'll perform few test runs in July to see if results will get substantially better when increasing the transfer penalty for leisure trips.

Scripts/assignment/emme_assignment.py
```diff
@@ -332,7 +340,7 @@ def print_vehicle_kms(self, resultdata):
             transit_times.values(), "transit_kms.txt",
             transit_times.keys(), "time")
 
-    def calc_transit_cost(self, fares, peripheral_cost, default_cost=None):
+    def calc_transit_cost(self, transit_class, fares, peripheral_cost, default_cost=None):
```
zptro, line 343: Parameter `transit_class` is overwritten in for-loop, so it does not have any effect.

Scripts/modelsystem.py
```diff
@@ -97,9 +97,10 @@ def assign_base_demand(self, use_fixed_transit_cost=False):
                 with self.basematrices.open("cost", "aht") as aht_mtx:
                     fixed_cost = aht_mtx["transit"]
             else:
-                self.logger.info("Calculating transit cost")
                 fixed_cost = None
-            self.ass_model.calc_transit_cost(self.zdata_forecast.transit_zone, peripheral_cost, fixed_cost)
+            for transit_class in ["transit_work", "transit_leisure"]:
+                self.logger.info("Calculating transit cost for class {}".format(transit_class))
+                self.ass_model.calc_transit_cost(transit_class, self.zdata_forecast.transit_zone, peripheral_cost, fixed_cost)
```
zptro, line 103: I think this is overkill. Path choices probably do not vary that much over assignment classes that it affect transit fares in any substantial way. Takes long time to run. We could use the same matrix in both classes.

Scripts/assignment/emme_assignment.py
```diff
@@ -724,46 +673,178 @@ def _calc_extra_wait_time(self, scen_id):
...
+                tp_attr[segment.id] = segment[extra_attr]
+            segments_attr[tp] = tp_attr
+        # save link volumes to result network
+        day_scenario = emmebank.scenario(self.day_scenario)
+        network = day_scenario.get_network()
+        for segment in network.transit_segments():
+            day_attr = 0
+            for tp in self.emme_scenarios:
+                if segment.id in segments_attr[tp]:
+                    day_attr += segments_attr[tp][segment.id] * param.volume_factors["transit"][tp]
```
zptro, line 769: Need to use the class-specific volume factor

zptro (changes requested):
Looking forward to seeing results from this one! But first we need to fix some things:

Need to update test matrices to include new assignment classes.

Need to update this with new parameter or remove the parameter altogether: https://github.com/HSLdevcom/helmet-model-system/blob/9d7ec8ac68b4579b086248dc05e876293d0bcbbf/Scripts/assignment/mock_assignment.py#L86

Scripts/assignment/emme_assignment.py
zptro: https://github.com/HSLdevcom/helmet-model-system/pull/183/commits/4b645a943f21dc623b5eb6f558222759a203acd2

Scripts/modelsystem.py
zptro: https://github.com/HSLdevcom/helmet-model-system/pull/183/commits/4b645a943f21dc623b5eb6f558222759a203acd2

Scripts/assignment/emme_assignment.py
zptro: https://github.com/HSLdevcom/helmet-model-system/pull/183/commits/70c6f395516fb2ed37dcf23108fee46d36f0525d

zptro (approved):

PR #184, opened 2020-08-03
Make running model with mock assignment independent of emme library

zptro:
Make running model with mock assignment independent of emme library

Also added more docstrings and improved readability of helmet.py


TimoElolahde:
Good improvements

PR #187, opened 2020-08-12
Parking norm possible to define manually

zptro:
Added possibility to have a .car file for forecast years (normally this file would exist only for base year and contain true car density) containing parking norms (unit: cars per inhabitant) in column _prknorm_. The file can contain values for a subset of the zones. 

The parking norm, when defined, will override modelled car densities for new inhabitants. Zones that are missing from the file will adopt modelled car densities.

Also made some other smaller adjustments and fixed bug in type casting in `ZoneData`.

### DIFF
--- a/Scripts/assignment/emme_assignment.py
+++ b/Scripts/assignment/emme_assignment.py
@@ -339,13 +339,27 @@ class EmmeAssignmentModel(AssignmentModel, ImpedanceSource):
             transit_times.keys(), "time")
 
     def calc_transit_cost(self, fares, peripheral_cost, default_cost=None):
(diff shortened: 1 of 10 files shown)


TimoElolahde (approved):

zptro:
Thanks for the comment, @atte-wsp! I think the problem occurs in base data for zone 1127.

zptro:
> 
> 
> Should we divide with absolute `pop_growth` instead of `pop_growth_share` to get share of new dwellings that are detached houses? Now `"share_detached_houses_new"` can be negative and over one, which gives an error in `ShareChecker`.

Yes, you are right, I will fix this.

PR #190, opened 2020-08-25
Fix share of detached houses of new dwellings

zptro:
https://github.com/HSLdevcom/helmet-model-system/pull/187#commitcomment-41750843

### DIFF
--- a/Scripts/models/linear.py
+++ b/Scripts/models/linear.py
@@ -87,8 +87,8 @@ class CarDensityModel(LinearModel):
         detached_houses_diff = (forecast_sh_detached*forecast_pop 
                                 - base_sh_detached*base_pop)
         share_detached_new = numpy.divide(
-            detached_houses_diff, self.pop_growth_share,
-            out=numpy.array(forecast_sh_detached), where=self.pop_growth_share!=0)
+            detached_houses_diff, pop_growth,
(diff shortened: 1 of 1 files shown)


atte-wsp (changes requested):
This still throws error. Some zones in MAL2019 forecasts (for example id 1051) have substantial growth in share of detached houses even if population growth is close to zero. In these cases `share_detached_houses_new` can get values greater than one. 

I think here the error is in input data, which should not have changes in share of detached houses if population stays same.

zptro:
> 
> 
> This still throws error. Some zones in MAL2019 forecasts (for example id 1051) have substantial growth in share of detached houses even if population growth is close to zero. In these cases `share_detached_houses_new` can get values greater than one.
> 
> I think here the error is in input data, which should not have changes in share of detached houses if population stays same.

Thank you for observing this! I think we will have to allow these kind of input data, because in some cases land development might involve destruction of existing buildings. Even if some detached houses are replaced by more dense development, population density in existing buildings often decreases gradually. There is no easy way to model this correctly in the current framework, but I think allowing `share_detached_houses_new` to go below zero or above one wuold be an ok approximation.

atte-wsp:
I agree on previous reasoning.

atte-wsp (approved):

PR #192, opened 2020-09-03
Save Network Results after Congested Transit Assignment

atte-wsp:
* Save results to extra attribute after last assignment
* Fixes problem with 24h expansion of transit volumes, where current results were different than reported earlier


zptro:
Already resolved in https://github.com/HSLdevcom/helmet-model-system/pull/183

PR #202, opened 2020-09-21
Transit generation calibration

atte-wsp:
* Add transit generation dummy-variables to correct generation of transit trips in different areas (cbd, helsinki_other, espoo_vant_kau).
* Mode share calibration was not included, so mode shares need to be calibrated later. This is done later after new cost estimation.

Calibration file: https://hslfi.sharepoint.com/:x:/r/sites/ext-helmet_kehi_40/Jaetut%20asiakirjat/6%20Analyysit/Kalibrointi/Vertailut/calibration_transit_dummy_dest_v01.xlsx?d=w9075b6758179418fac1686562898108e&csf=1&web=1&e=S8mHkg

### DIFF
--- a/Scripts/datahandling/zonedata.py
+++ b/Scripts/datahandling/zonedata.py
@@ -90,6 +90,8 @@ class ZoneData:
         self.share["share_detached_houses"] = landdata["detach"]
         self["cbd"] = pandas.Series(0, self.zone_numbers)
         self["cbd"].loc[:param.areas["helsinki_cbd"][1]] = 1
+        self["helsinki_other"] = pandas.Series(0, self.zone_numbers)
+        self["helsinki_other"].loc[:param.areas["helsinki_other"][1]] = 1
         self["helsinki"] = pandas.Series(0, self.zone_numbers)
         self["helsinki"].loc[:param.municipality["Helsinki"][1]] = 1
         self["espoo_vant_kau"] = pandas.Series(0, self.zone_numbers)
--- a/Scripts/datahandling/zonedata.py
+++ b/Scripts/datahandling/zonedata.py
@@ -89,15 +89,15 @@ class ZoneData:
         self["zone_area"] = landdata["builtar"]
         self.share["share_detached_houses"] = landdata["detach"]
         self["cbd"] = pandas.Series(0, self.zone_numbers)
-        self["cbd"].loc[:param.areas["helsinki_cbd"][1]] = 1
+        self["cbd"].loc[param.areas["helsinki_cbd"][0]:param.areas["helsinki_cbd"][1]] = 1
         self["helsinki_other"] = pandas.Series(0, self.zone_numbers)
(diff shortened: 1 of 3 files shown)


atte-wsp:
Some details of trip destinations:
* hc and hs trips are too long. For both groups there are too many trips from helsinki_other to helsinki_cbd and too few from helsinki_other to helsinki_other.

I'll see if these can be calibrated with current variables.

Scripts/datahandling/zonedata.py
```diff
@@ -90,6 +90,8 @@ def __init__(self, data_dir, zone_numbers):
         self.share["share_detached_houses"] = landdata["detach"]
         self["cbd"] = pandas.Series(0, self.zone_numbers)
         self["cbd"].loc[:param.areas["helsinki_cbd"][1]] = 1
+        self["helsinki_other"] = pandas.Series(0, self.zone_numbers)
+        self["helsinki_other"].loc[:param.areas["helsinki_other"][1]] = 1
```
zptro, line 94: I realize now that I have duplicated the "cbd"-dummy row (for "espoo_vant_kau" and "surrounding") without further thinking, and so has @atte-wsp done as well. We now do `.loc[0:upper]` when we should do `.loc[lower:upper]`.

Fortunately the "espoo_vant_kau" has not been used anywhere and "surrounding" dummy affects only the car density model.

zptro (changes requested):

zptro:
> 
> 
> Some details of trip destinations:
> 
>     * hc and hs trips are too long. For both groups there are too many trips from helsinki_other to helsinki_cbd and too few from helsinki_other to helsinki_other.
> 
> 
> I'll see if these can be calibrated with current variables.

I guess this can be re-calibrated using the "cbd" dummy in destination choice.

Scripts/parameters.py
```diff
@@ -1863,7 +1863,11 @@
         "transit": {
             "constant": (0.844179295926 * (3.15335645065 - 0.158), 
                          0.844179295926 * (3.09257722177 - 0.415)),
-            "generation": {},
+            "generation": {
+                "cbd": 0.3,
+                "helsinki_other": 0.3,
+                "espoo_vant_kau": -0.2,
+            },
```
zptro, line 1870: I think that if the bug mentioned above is fixed, the dummy parameters can be easily modified like this:
`"cbd": 0.3+0.3-0.2,
"helsinki_other": 0.3-0.2,
"espoo_vant_kau": -0.2,`
because I think that is the effect right now.

atte-wsp: True, need to try this first

Scripts/datahandling/zonedata.py
atte-wsp: And I was even thinking that variables of cbd and helsinki_other seemed to be oddly correlated! Never should use copy / paste when coding. Luckily this was easy to fix.

zptro (approved):

PR #204, opened 2020-09-23
End assignment calibration

zptro:
Separated boarding penalties and transfer penalties for end assignment and regular assignment. 

In end assignment:
- transfer penalty for work trips 5 -> 3
- boarding penalty for buses 3 -> 5
- boarding penalty for trunk buses 3 -> 2


zptro:
@atte-wsp, maybe you could try this now with the re-calibrated demand model results?

atte-wsp:
I'll try these changes. It will take couple of hours to run end assignments, but results should be ready today.

atte-wsp:
I think this looks quite good. However, the average number of boardings per trip is still 1.52, which is really high. Should we try to use transfer penalty 5 / 7 minutes?

image

zptro:
> 
> 
> I think this looks quite good. However, the average number of boardings per trip is still 1.38, which is a little too much. Should we try to use transfer penalty 5 / 7 minutes?
> 
> image
> 
> Edit. I had calculated the avg number of boarding wrongly. Now the calc is `number of boardings (hsl modes) / transit trips (pks & surrounding)`

I think this is ok. We should have a bit too many boardings per trip to make up for missing secondary-destinations (third, fourth, etc.)

atte-wsp (approved):

PR #209, opened 2020-10-06
Update 4verkkoa_HM40.mac

TimoElolahde:
Creating extra attributes must be done when creating new Emme scenarios in the Emme project. When created once new values can be read or modified as many times as needed. If the extra attributes exist already recreating them causes an error message. So the line is modified to a comment.

### DIFF
--- a/Database/4verkkoa_HM40.mac
+++ b/Database/4verkkoa_HM40.mac
@@ -157,7 +157,7 @@ reports=
  1
  y
  q
-~< extra_attr.mac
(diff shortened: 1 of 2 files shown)


zptro (approved):
I re-added the `aja_extra_attr_HM31.mac` macro, so that the suggested method can be followed. Also updated instructions: https://github.com/HSLdevcom/helmet-ui/commit/10dccc11452fd662fe10baa477e98c45b5f277f8

PR #212, opened 2020-10-06
Update 4verkkoa_HM40.mac

TimoElolahde:
It looks like Atte’s philosophy has been that in all scenarios extra attribute table is first initialized and then the ordinary set of extra attributes needed in the model system is recreated. However, the user may have defined also his/her own extra attributes which are lost when using this method. That’s why in my opinion it’s better to define the extra attributes already when creating a new Emme project. Then the attributes exist, and new values can be read as many times as needed. If the user wants to add or delete his/her own attributes it’s possible.

If the user still wants to define the extra attributes in macro 4verkkoa_M2019.mac it must be done before calling the macro muuta_linkkien_attribuutit_eikorj.mac.

In comparison to the nodes, links and transit lines tables, there is no need to do partial modifications. It’s better to initialize the tables and reread the complete nodes, links and transit lines data.

### DIFF
--- a/Database/4verkkoa_HM40.mac
+++ b/Database/4verkkoa_HM40.mac
@@ -65,11 +65,11 @@ reports=
  2
  off=11
 ~# tyhjenna ja lue pyoraverkon extra attribuutit
- 2.42
- 1
- y
- q
-~< extra_attr_pyora.mac
+~#  2.42
+~#  1
+~#  y
+~#  q
+~# ~< extra_attr_pyora.mac
 ~# ** luetaan pyoratieluokat (makro poistaa tiedoston d241.in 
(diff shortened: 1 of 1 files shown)


zptro:
I cannot figure out why did does not pass the tests

> UnicodeDecodeError: 'utf8' codec can't decode byte 0x92 in position 44: invalid start byte

zptro:
> 
> 
> I cannot figure out why did does not pass the tests
> 
> > UnicodeDecodeError: 'utf8' codec can't decode byte 0x92 in position 44: invalid start byte

It seems commit messages should not contain the *right single quotation mark* `’`. @TimoElolahde, use `'` instead!

zptro (approved):

PR #213, opened 2020-10-07
Update Results Aggregation

atte-wsp:
* Call 24h expansion only once for each class and attribute
* More informative log messages

### DIFF
--- a/Scripts/modelsystem.py
+++ b/Scripts/modelsystem.py
@@ -191,6 +191,13 @@ class ModelSystem:
             if is_end_assignment:
                 self._save_to_omx(impedance[tp], tp)
         if is_end_assignment:
+            for ass_class in parameters.link_volumes:
+                self.ass_model.auto_link_24h(ass_class)
+            for transit_class in parameters.transit_classes:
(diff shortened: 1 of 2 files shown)


Scripts/modelsystem.py
```diff
@@ -191,6 +191,13 @@ def assign_base_demand(self, use_fixed_transit_cost=False, is_end_assignment=Fal
             if is_end_assignment:
                 self._save_to_omx(impedance[tp], tp)
         if is_end_assignment:
+            for ass_class in parameters.link_volumes:
+                self.ass_model.auto_link_24h(ass_class)
+            for transit_class in parameters.transit_classes:
+                self.ass_model.transit_segment_24h(transit_class, "vol")
+                self.ass_model.transit_segment_24h(transit_class, "boa")
+                self.ass_model.transit_segment_24h(transit_class, "trb")
+            self.ass_model.bike_link_24h()
```
zptro, line 200: Maybe bundle these into one function in EmmeAssignment? Could maybe be in the same function as vehicle km prints.

zptro (approved):

PR #218, opened 2020-10-20
In loop, add transit matrices to transit_work matrix

zptro:
In previous optimization, transit_leisure assignment was removed from demand calculation loop as it is redundant in skim matrix calculation. However, if the model run is for some reason cut short (i.e., congested assignment not performed) the assignment results should be correct anyway.

Now transit_leisure demand is added to transit_work, so that sum in assignment result will be correct.


zptro:
@atte-wsp, I am planning to merge this today (running full scale tests right now). If you do not have any comments, could you approve?

atte-wsp (approved):
For some reason I thought that I already approved this. There are many sequential if clauses, but they seem to be necessary, or at least I could not find a way to make it easier to read. Working ok in any case.

PR #219, opened 2020-10-20
Moved parameters to own package and splitted into several modules

zptro:
Organized parameters according to topic into several submodules. The correct submodule is in general imported as `param`.


TimoElolahde (approved):
Changes seem to be reasonable.

atte-wsp (approved):
Easier to read and more explicit than before.  Should we have different names for different parameter imports (as there is `zone_param` and `generation_param` in `logit.py`)?

zptro:
> 
> 
> Easier to read and more explicit than before. Should we have different names for different parameter imports (as there is `zone_param` and `generation_param` in `logit.py`)?

Will consider this for future updates. Will merge this one now.

PR #220, opened 2020-10-20
Fix hop parameters

atte-wsp:
* Trip purpose "hop" destination choice parameters had wrong coefficients for impedance. Error in latest update.
* These correspond to Helmet-estimation files and model report


zptro (approved):
Good that you found this!

atte-wsp:
I'll still check the calibration, since this could affect those parameters as well.

TimoElolahde (approved):
Fine

PR #221, opened 2020-10-20
Add Aggregate Results to Mock Assignment

atte-wsp:
* Mock Assignment model run crashes due to this method is not found
* `AttributeError: 'MockAssignmentModel' object has no attribute 'aggregate_results'`


zptro (approved):
Also fixed in https://github.com/HSLdevcom/helmet-model-system/pull/219, but probably better to have as separate pull.

PR #226, opened 2020-10-26
Updated volume expansion factors in cba excel

zptro:
Updated factors for volume expansion to whole day
Removed faulty division by 60 for distances (error in previous versions!)


TimoElolahde (approved):
OK

PR #230, opened 2020-11-03
Base matrix check

zptro:
Check that input matrices have the correct dimensions (and zone numbers in mapping). If zones are missing from matrices compared to network, zero rows and columns are inserted. If matrices have extra zones, an error is thrown in indata validation.


TimoElolahde (approved):
Sounds reasonable and looks fine!

samulikyytsonen (approved):
Looks really good!

PR #231, opened 2020-11-17
Refactor logging and area aggregation

zptro:
Wrapping the singleton logging with a class has been confusing. Now this is simplified so that logging is wrapped pythonically with static functions. Added and clearified some loggings.

Also streamlined slicing according to zone intervals from parameters (municipalities or areas) into own utility and did some improvements of area aggregations of output results.

Updated some docstrings.

### DIFF
--- a/Scripts/assignment/departure_time.py
+++ b/Scripts/assignment/departure_time.py
@@ -1,6 +1,6 @@
-import logging
 import numpy
 
+import utils.log as log
 import parameters.departure_time as param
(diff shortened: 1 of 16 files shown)


TimoElolahde (approved):
Sounds reasonable.

atte-wsp:
Running with Emme crashes:
> AttributeError: 'RootLogger' object has no attribute 'add_handler'

Didn't yet figure out what is the cause, but I should have version corresponding the master.

PR #235, opened 2020-11-18
Fix Emme Project logging

atte-wsp:
* Unsure how the logging is working, but I think that `add_handler` should be `addHandler`


zptro (approved):
Thanks for noticing this!

PR #237, opened 2020-11-23
Save impedances for MockAssignment

atte-wsp:
Not sure if this should be part of master. I think it would make our work easier at beginning, when we need to distinguish the effects of transit congestion, but this might be unnecessary feature for most users.
* Print omx-matrices with impedances from iteration previous to final round.
* This allows using MockAssignment (impedances written to file are same that are used in mode choice) and it is updated to account changes in file names.
* Allows to study the effects of congestion in transit impedances, which might be important to distinguish in CBA and impact analysis.
* Disadvantage is that omx-files will consume more hard disk space (around ~~200~~ 20 MB).


Scripts/modelsystem.py
```diff
@@ -279,6 +279,7 @@ def run_iteration(self, previous_iter_impedance, iteration=None):
             if tp == "aht":
                 self._update_ratios(impedance[tp], tp)
             if iteration=="last":
+                impedance[tp]["std_time"] = previous_iter_impedance[tp]["time"]
```
zptro, line 282: This consumes a lot of hard disk space, because we store copies of travel time matrices for all modes. What if we would save uncongested transit times in `impedance[tp]["time"]["transit_uncongested"]` instead? Would then need to modify `MockAssignment` accordingly of course.

zptro (commented):
Need to update test data and merge master changes in MatrixData.

zptro (approved):

zptro:
@atte-wsp, what do you think? Ready to merge?

atte-wsp (commented):
Seems that I cannot approve as I'm author of this Pull Request. However, I checked the changes and would approve. 

Works much better now that there are not that many extra matrices saved to file. However, seems that we have same benefits that I was looking for.

PR #238, opened 2020-11-25
Save emme matrices

### LINKED ISSUE #214: Transit result matrices are not saved
At the moment, only generalized (weighted) travel time matrices are saved from transit assignment. For some analyses, trip parts and unweighted travel time is useful as well.

zptro:
Added possibility to save all demand and result matrices in Emme format. This includes transit part travel time part matrices (in-vehicle time, etc.), which have not been stored anywhere so far. The numbering of matrices will be 100+ (compared to old matrix numbering) for aht, 200+ for pt and 300+ for iht by default, but the starting number can be changed in config-file.

Closes  #214.


atte-wsp (approved):
Good update! This enables also extended transit analysis after Helmet-run. 

I was still thinking if starting point of saved demand matrices could be defined by user? Most use cases require us to compare Helmet-scenarios and therefore it would be beneficial to not to overwrite demand matrices and extended assignment analysis. However this could be too complicated.

In addition could remove extra matrices from parameters. If I'm correct the separate bike matrices were used only in development phase.

zptro:
> 
> 
> Good update! This enables also extended transit analysis after Helmet-run.
> 
> I was still thinking if starting point of saved demand matrices could be defined by user? Most use cases require us to compare Helmet-scenarios and therefore it would be beneficial to not to overwrite demand matrices and extended assignment analysis. However this could be too complicated.
> 
> In addition could remove extra matrices from parameters. If I'm correct the separate bike matrices were used only in development phase.

Thanks for the feedback! Now the first matrix id can be changed in config. When we run multiple scenarios in batch from GUI, they must have the same matrix ids, but batch runs with different matrix ids can be performed from command line.

Also removed unnecessary bike matrices.

atte-wsp (approved):
Perfect!

PR #241, opened 2020-11-30
CBA output

atte-wsp:
Wanted to get time gains by origins, ended up refactoring cba.py quite much. 

CBA
* Calculates 24h benefits for all items using param.volume_factors
* Handles gains and revenues in matrix shape and prints out gains by origins
* Include possibility to print travel costs and gains to selected destinations 

This needs still some work, but I think it could be useful for others as well.


atte-wsp:
I'll close Pull Request as we're modifying CBA for Viikki-Malmi light rail feasibility study. However, printing gains by origins and calculation in matrix form could be useful in some phase of model development and analysis.

Hopefully CBA interface to GUI and Helmet-model results is not heavily modified, so we can continue using same code.

zptro:
Ok, good to know, @atte-wsp. I was actually pondering what to do with this one, I was at least somewhat positive about merging it with master, but I haven't had time to go through it in detail.

Planned development of CBA calculation:
- Inclusion of network based noise calculation. This will be a small output file from model run, a bit similar to for instance vehicle-kms.txt.
- Updates to accident prevalence and pollution per km in the excel file.

atte-wsp:
@zptro Some of these changes could still be useful for CBA calculations and result outputs. I've used the gains by origins to check that user benefits are in line with changes in network. Writing to CBA-kehikko Excel-file is now detached in this branch, but that is quite easy to fix if necessary.

Also: Noticed that in CBA_kehikko discount calculations have `verokerroin` applied to `Liikennöintikustannus`. I'm not an expert on this economic part, but it seems that [Väylä p. 52](https://julkaisut.vayla.fi/pdf11/vo_2020-39_ratahankkeiden_arviointiohje_web.pdf) states that it should be applied only to `investointi` and `kunnossapito`.

PR #249, opened 2021-01-11
Aggregate peripheral cost matrix

zptro:
Transit cost (monthly card) data for peripheral area is aggregated to municipality level. Matkahuolto distance-based cost (if no monthly card is available) will be calculated for this area now as well, if value in file is 999 (or more).

Backwards-compatibility with old data format is maintained. If new-format file is not found, old format is used.

This will allow zones to be added without hassle.


TimoElolahde (approved):
Sounds reasonable and useful

samulikyytsonen (approved):
Looks great! Thanks again for making additional zone forecasts easier!

PR #255, opened 2021-01-22
Refactor Emme-assignment 2.0

zptro:
This PR borrows from #165, but is built on the concept of an `AssignmentPeriod` instead of  *Emme scenario*. So bike assignment does not have an object of its own, it just refers to another Emme scenario number which is saved in the same `AssignmentPeriod` as the other modes. Methods for aggregating to the whole day are directly under `EmmeAssignmentModel`.


atte-wsp (approved):
Works ok!

PR #251, opened 2021-01-13
new versions of input macros

zptro:
Cherry-picked commit from #245 (which is a pull request to branch https://github.com/HSLdevcom/helmet-model-system/tree/TimoElolahde-patch-2).

This will merge directly to master.

PR #256, opened 2021-02-01
Fix bugs in network import

zptro:
The additions to input macros in #251 uses `@ccost` as a temp variable, but this attribute was not added in the macros (it is created in the model run).  I also messed up the merge a little bit, so this is fixed now.


TimoElolahde (approved):
OK

PR #257, opened 2021-02-01
Remove file-specific log messages

### LINKED ISSUE #254: Strategy file removals flood log
We should print just one message about file removals.

zptro:
- Remove file-specific log messages
- Restructure path-joins

Closes #254


atte-wsp (approved):
Nice! Logging much better now as the was far too many logging msgs.

PR #265, opened 2021-02-22
Calculate work-force accessibility and "SAVU"

zptro:
Work-force accessibility is calculated by inverting work tour model so that it calculates population accessibility from the workplaces' point of view. Had to drop a number of variables connected to the residential zone, because they cannot easily be transferred to the attraction side of the model and vice versa. Not using only 18-64 population, because we lack that information for peripheral area.

Ideally, we should estimate own parameters for this model and not just reuse them from the work tour model.

Accessibility using sustainable modes (transit, bike, walk) is now printed to separate file. These are used to classify zones into seven [*SAVU*](https://www.hsl.fi/sites/default/files/uploads/saatavuustarkastelut_savu_hlj2011_raportti.pdf) classes.


Scripts/parameters/zone.py
```diff
@@ -138,4 +144,5 @@
         "surrounding",
         "peripheral",
     )
+SAVU_intervals = [6, 5, 4, 3, 2, 1, -9999] # TODO Calculate intervals
```
zptro, line 147: @johpiip, this PR is now done, except that I need to calculate the values that should be inserted here

Scripts/parameters/zone.py
```diff
@@ -93,6 +93,17 @@
...
+        (6000, 6999),
+        (10000, 11999),
+        (13000, 14999),
+        (15500, 15999),
+    ),
+    "surround_other": (
+        (7000, 9999),
+        (12000, 12999),
+        (15000, 15499),
+    ),
```
zptro, line 106: Unfortunately the area definition used in Ramboll indicators is much more complicated and requires substantial code changes in `ZoneIntervals`.

johpiip: I don't understand! Can you elaborate here or in chat?

johpiip (approved):

PR #266, opened 2021-03-05
Agents output

atte-wsp:
Methods to print accessibility values and other results from agent model. 

### Accessibility
Accessibility defined here as expected maximum utility from mode choice model, which is the denominator of the logit choice model (`logsum`). Mode specific logsums are also printed for `car` and `other.` 

Logsum is calculated for each agent and tour in model after individual dummies are added. Tour utilities are summed for each agent.

### Scaling of accessibility
The accessibility is in the unit of utils. Utils can be compared within tour purposes, but not however sure if this is case between groups, as each group has a different perception for the value of one unit of utils. 

Dong et al. 2006 argue that accessibility measure should satisfy two conditions:
1) The scale condition requires all the calculated accessibility values to share the same scale or same unit.
2) The level condition requires the values to have a consistent benchmark utility. This is necessary because the underlying model
does not depend on the absolute size of utility but on the utility differences of the available alternatives. Any constant could be added to all alternatives utilities without changing the choice model probabilities

I would interpret these so that logsum between tour purposes (different model specification) cannot be compared, because the level is arbitrary and does not describe relative importance of trip.

Scaling of accessibility is a bit tricky since all modes do not share common cost or time parameter. Needs further investigations:

Sources:
[Dong et al. 2006](https://xjdong.github.io/papers/2006_TRA_Activity%20Accessibility.pdf)
[Miller 2020](https://www.itf-oecd.org/sites/default/files/docs/measuring-accessibility-methods-issues.pdf)
[Xi 2019](https://tspace.library.utoronto.ca/bitstream/1807/98452/3/Xi_Yang_%20_201911_MAS_thesis.pdf)


zptro:
> Dong et al. 2006 argue that accessibility measure should satisfy two conditions:
> 
> 1. The scale condition requires all the calculated accessibility values to share the same scale or same unit.
> 2. The level condition requires the values to have a consistent benchmark utility. This is necessary because the underlying model does not depend on the absolute size of utility but on the utility differences of the available alternatives. Any constant could be added to all alternatives utilities without changing the choice model probabilities
>
> I would interpret these so that logsum between tour purposes (different model specification) cannot be compared, because the level is arbitrary and does not describe relative importance of trip.
>
> Scaling of accessibility is a bit tricky since all modes do not share common cost or time parameter.

These are good remarks. 

1. The scale can be unified by translating utility into money or time. This does not require the parameter to be the same for all purposes. I would advocate for using the money unit, as marginal utility of time naturally varies between tour purposes but marginal utility of money should not depend on the tour undertaken. A problem is school tours, which do not have a parameter for linear cost. We *can* translate this utility into money from exp(utility), but here I have used a more simple approach: https://github.com/HSLdevcom/helmet-model-system/blob/0f109d569080dc00abfd6fd8cd54c5bd25f5a7d7/Scripts/models/logit.py#L73 I am actually using time unit for school tours, but translating it into money using CBA guidelines.
2. It is correct that logsums from different tour purposes cannot be compared, but at least they can be summed (using weight based on tour prevalence), and logsum sums from different scenarios can be compared.

zptro:
Squash merges fuck up version control. Sorry about this.

PR #267, opened 2021-03-08
Add Zone class

zptro:
I have thought about this for a long time, but it has not really made any sense before. But now as `Person` needs area and municipality membership information for the home zone, I thought it was time to create a `Zone` class, which stores information on area and municipality membership and zone index.

### DIFF
--- a/Scripts/datatypes/person.py
+++ b/Scripts/datatypes/person.py
@@ -12,8 +12,8 @@ class Person:
     
     Parameters
     ----------
-    zone : int
-        Zone number, where person resides
+    zone : datatypes.zone.Zone
+        Zone where person resides
     age_group : tuple
         int
             Age interval to which the person belongs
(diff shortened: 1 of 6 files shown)


Scripts/datatypes/person.py
```diff
@@ -61,21 +61,11 @@ def calc_income(self):
...
 
     @property
     def municipality(self):
-        result = None
-        for mp in municipalities:
-            if  municipalities[mp][0] <= self.zone < municipalities[mp][1]:
-                result = mp
-                break
-        return result
+        return self.zone.municipality
```
zptro, line 68: References to these can be replaced altogether by `self.zone.area` and `self.zone.municipality`, but I kept them for now to not break existing code.

atte-wsp (approved):
Good idea and solution works nicely. I think we could we merge this to #266 right away and I can then replace the unnecessary property methods.

PR #269, opened 2021-03-10
Vehicle kms and noise calculation

zptro:
- Sum vehicle kilometres to area level
- Calculate noise areas
- Print numbers of metro and train stations (for use in CBA maintenance cost calculation)
- Save car time matrices in Emme


Scripts/assignment/emme_assignment.py
```diff
@@ -245,6 +254,68 @@ def _create_attributes(self, scenario, attributes):
...
+        for link in network.links():
+            # Aggregate traffic
+            light_modes = ("@car_work", "@car_leisure", "@van")
+            traffic = sum([link[mode] for mode in light_modes])
+            rlink = link.reverse_link
+            if rlink is None:
+                reverse_traffic = 0
+            else:
+                reverse_traffic = sum([rlink[mode] for mode in light_modes])
+            cross_traffic = 0.85 * 0.9 * (traffic+reverse_traffic)
```
johpiip, line 270: `calc_noise()` contains several contants, like 0.85 and 0.9 here, and their purposes are unclear for me. :) Consider making them variables with descriptive names ie. `variable_name_telling_what_the_number_means = 0.85` or adding some other comment. Apart from this row, magic numbers appear also in many other lines in this function.

johpiip (changes requested):
Nice! I added a couple of comments regarding documentations. One other issue though... Are we still using `pylint`? The linter throws 215 problems in these changed files (trailing whitespace, too long lines, missing docstrings, naming conventions...). We could also do a separate PR for cleaning those here and in other scripts.

Scripts/parameters/zone.py
```diff
@@ -148,13 +148,62 @@
...
+    "helsinki_cbd",
+    "helsinki_other",
+    "espoo_vant_kau",
+    "surround_train",
+    "surround_other",
+    "peripheral",
+)
 savu_intervals = (-175.2, -171.3, -166.56, -130.1, -123.15, -118.13, 9999)
 tour_length_intervals = (0, 1, 3, 5, 10, 20, 30, 40, float("inf"))
+pop_share_per_noise_area = {
```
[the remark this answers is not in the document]
zptro: Good point. Added.

Scripts/assignment/emme_assignment.py
zptro: These (0.85 and 0.9) were clear to me, so I made the suggested changes. However, other parameters are truly "magic" in the sense that I do not know the meaning of them. I am not sure how to handle them.

zptro:
> Are we still using `pylint`?

The question is rather "Are we still *not* using `pylint`?" :smile:

A separate PR for fixing all the bad coding is a good idea.

Scripts/assignment/emme_assignment.py
johpiip: Good changes! Hmm, the second part is tricky! Is it possible to add a comment pointing to the source of those numbers (like a hyperlink if the source is a report or a website) even if the source does not say anything specific of the numbers? At least, there would be a clue about how to investigate them later. This is optional!

johpiip (approved):
Thanks for the changes! Looks good to me!

johpiip (approved):

zptro:
@johpiip, maybe this is a better way of representing the step-wise function?

Scripts/assignment/emme_assignment.py
```diff
@@ -245,6 +254,65 @@ def _create_attributes(self, scenario, attributes):
...
+                                    + 500*heavy_share/speed))
+            else:
+                heavy_correction = (10*log10((1-heavy_share)
+                                    + 5.6*heavy_share*(90/speed)**3))
+            start_noise = (68 + 30*log10(speed/50)
+                           + 10*log10(cross_traffic/15/1000)
+                           + heavy_correction)
+
+            # Calculate noise zone width
+            for i in param.noise_zone_width:
```
johpiip, line 305: (optional, highly personal preference) For me, `i` sounds like `integer` (but I guess some might interpret `i` as `item`, too). I'd prefer: `for noize_zone in param.noise_zone_width:`

Scripts/parameters/assignment.py
```diff
@@ -186,6 +186,21 @@
         "iht": 1 / 0.497,
     },
 }
+# Factor for converting weekday traffic into yearly day average
+years_average_day_factor = 0.85
+# Factor for converting day traffic into 7:00-22:00 traffic
+share_7_22_of_day = 0.9
+# Noice zone width as function of start noise
+noise_zone_width = [
+    {"threshold": 80, "function": lambda x: 225},
```
johpiip, line 195: What do you think about changing `threshold` to explicit `lower` and `upper` boundaries? Instead of `start_noise >= i["threshold"]` in https://github.com/HSLdevcom/helmet-model-system/pull/269/files#diff-2cffa1ae64c4860a3669589e445fcb62e50cda3379c945f94887344fd06f1e12R306 we would have mathematically "noise in [lower, upper)" so something like `noise >= i["lower"] & noise < i["upper"]`. I think it would make debugging easier.

Scripts/parameters/assignment.py
```diff
@@ -186,6 +186,21 @@
...
+    {"threshold": 77, "function": lambda x: 153 + 66/3*x},
+    {"threshold": 74, "function": lambda x: 109 + 44/3*x},
+    {"threshold": 71, "function": lambda x: 78 + 31/3*x},
+    {"threshold": 68, "function": lambda x: 57 + 21/3*x},
+    {"threshold": 65, "function": lambda x: 41 + 16/3*x},
+    {"threshold": 55, "function": lambda x: 10 + 31/10*x},
```
johpiip, line 201: Overall very nice refactoring here and in `calc_noise()` where these `function`s are being called! `lambda`s were a new thing for me, thanks! 😄

johpiip (approved):
Just a couple optional changes. Nice refactoring!

Scripts/parameters/assignment.py
zptro: Did something like this, hope you like it!

johpiip (approved):
Looks good to me!

PR #274, opened 2021-03-20
Agent model results

atte-wsp:
## Methods to print accessibility values and other results from agent model.
Replaces closed PR #266 as I had some serious difficulties with version control.

## Accessibility
Accessibility defined here as expected maximum utility from mode choice model, which is the denominator of the logit choice model (logsum). Logsum is calculated for each agent and tour in model after individual dummies are added. Tour utilities are summed for each agent separately for car, sustainable modes and all modes.

### Scaling of accessibility
[Dong et al. 2006](https://xjdong.github.io/papers/2006_TRA_Activity%20Accessibility.pdf) and [Miller 2020](https://www.itf-oecd.org/sites/default/files/docs/measuring-accessibility-methods-issues_1.pdf) argue that accessibility measure should satisfy two conditions:

1. The scale condition requires all the calculated accessibility values to share the same scale or same unit.
2. The level condition requires the values to have a consistent benchmark utility. This is necessary because the underlying model
does not depend on the absolute size of utility but on the utility differences of the available alternatives. Any constant could be added to all alternatives utilities without changing the choice model probabilities
I would interpret these so that logsum between tour purposes (different model specification) cannot be compared, because the level is arbitrary and does not describe relative importance of trip.

Here following #265 accessibility scale is unified by translating utility into money or time, therefore condition first is satisfied. Second is still work in progress, as it would mean that summing accessibility for trip purposes won't reflect their importance to agents? However, accessibility values will be updated later. 

Additionally one should consider how marginal utility of money differs in income groups, as pointed out [here](https://link.springer.com/article/10.1007/s11116-017-9853-4).

### DIFF
--- a/Scripts/datatypes/person.py
+++ b/Scripts/datatypes/person.py
@@ -140,7 +140,7 @@ class Person:
             sustainable_access += tour.sustainable_accessibility
             car_access += tour.car_accessibility
         # print to file
-        persondata = "{:d}\t{:s}\t{:s}\t{!s}\t".format(
+        persondata = "{:d}\t{:s}\t{:s}\t{!s}\t{:.0f}".format(
             self.id, self.age_group, self.gender, self.is_car_user, self.income)
         zonedata = "{:d}\t{:s}\t{:s}".format(
             self.zone.number, self.zone.area, self.zone.municipality)
--- a/Scripts/datatypes/person.py
+++ b/Scripts/datatypes/person.py
@@ -36,9 +36,6 @@ class Person:
         self._im = income_model
         self._car_use_draw = random.random()
         self._tour_combination_draw = random.random()
-        self.total_access = 0
-        self.sustainable_access = 0
-        self.car_access = 0
 
     def decide_car_use(self):
         car_use_prob = self._cm.calc_individual_prob(
@@ -121,3 +118,31 @@ class Person:
                 if random.random() < non_home_prob:
                     non_home_tour = Tour(purposes["oo"], tour)
                     self.tours.append(non_home_tour)
+
+    def write_file(self, resultdata, filename):
+        """ Write person data to file.
+
+        Parameters
+        ----------
+        resultdata : ResultData
+            Writer object to result directory
+        """
+        # sum accessibility of all tours
+        nr_tours = 0
+        total_access = 0
+        sustainable_access = 0
+        car_access = 0
+        for tour in self.tours:
+            nr_tours += 1
+            total_access += tour.total_accessibility
+            sustainable_access += tour.sustainable_accessibility
+            car_access += tour.car_accessibility
+        # print to file
+        persondata = "{:s}\t{:s}\t{!s}\t".format(
+            self.age_group, self.gender, self.is_car_user, self.income)
+        zonedata = "{:d}\t{:s}\t{:s}".format(
+            self.zone.number, self.zone.area, self.zone.municipality)
+        accessdata = "{:d}\t{:.1f}\t{:.1f}\t{:.1f}".format(
+            nr_tours, total_access, sustainable_access, car_access)
+        resultdata.print_line("\t".join(
+            [persondata, zonedata, accessdata]), filename)
--- a/Scripts/datatypes/person.py
+++ b/Scripts/datatypes/person.py
@@ -132,12 +132,10 @@ class Person:
             Writer object to result directory
         """
         # sum accessibility of all tours
-        nr_tours = 0
         total_access = 0
         sustainable_access = 0
         car_access = 0
         for tour in self.tours:
-            nr_tours += 1
             total_access += tour.total_accessibility
             sustainable_access += tour.sustainable_accessibility
             car_access += tour.car_accessibility
@@ -147,6 +145,6 @@ class Person:
         zonedata = "{:d}\t{:s}\t{:s}".format(
             self.zone.number, self.zone.area, self.zone.municipality)
         accessdata = "{:d}\t{:.1f}\t{:.1f}\t{:.1f}".format(
-            nr_tours, total_access, sustainable_access, car_access)
+            len(self.tours), total_access, sustainable_access, car_access)
         resultdata.print_line("\t".join(
             [persondata, zonedata, accessdata]), filename)
--- a/Scripts/datatypes/person.py
+++ b/Scripts/datatypes/person.py
@@ -25,7 +25,9 @@ class Person:
     FEMALE = 0
     MALE = 1
     
-    def __init__(self, zone, age_group, generation_model, car_use_model, income_model):
+    def __init__(self, person_id, zone, age_group, 
+                 generation_model, car_use_model, income_model):
+        self.id = person_id
         self.zone = zone
         self.age = random.randint(age_group[0], age_group[1])
         self.age_group = "age_" + str(age_group[0]) + "-" + str(age_group[1])
@@ -138,8 +140,8 @@ class Person:
             sustainable_access += tour.sustainable_accessibility
             car_access += tour.car_accessibility
         # print to file
-        persondata = "{:s}\t{:s}\t{!s}\t".format(
-            self.age_group, self.gender, self.is_car_user, self.income)
+        persondata = "{:d}\t{:s}\t{:s}\t{!s}\t".format(
+            self.id, self.age_group, self.gender, self.is_car_user, self.income)
         zonedata = "{:d}\t{:s}\t{:s}".format(
             self.zone.number, self.zone.area, self.zone.municipality)
         accessdata = "{:d}\t{:.1f}\t{:.1f}\t{:.1f}".format(
--- a/Scripts/datatypes/person.py
+++ b/Scripts/datatypes/person.py
@@ -25,6 +25,10 @@ class Person:
     ID = 0
     FEMALE = 0
     MALE = 1
+    person_attr = ["id", "age_group", "gender", "is_car_user", "income", "nr_tours"]
+    zone_attr =  ["number", "area", "municipality"]
+    tour_attr = ["total_access", "sustainable_access", "car_access"]
+    attr = person_attr + zone_attr + tour_attr
     
     def __init__(self, zone, age_group, 
                  generation_model, car_use_model, income_model):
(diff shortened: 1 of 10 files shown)


atte-wsp:
@samulikyytsonen could you check this first and then ask review from Jens as well?

samulikyytsonen (dismissed):

Scripts/models/logit.py
```diff
@@ -426,10 +428,37 @@ def calc_individual_mode_prob(self, is_car_user, zone):
...
+        # utils to money
+        logsum = numpy.log(mode_expsum)
+        sust_logsum = numpy.log(sust_expsum)
+        car_logsum = numpy.log(car_expsum)
+        try:
+            b = self.dest_choice_param["car"]["impedance"]["cost"]
+        except KeyError:
+            # School tours do not have a constant cost parameter
+            # Use value of time conversion from CBA guidelines instead
+            b = -0.31690253
```
zptro, line 447: This code is now used in two places. Maybe we should put it in a separate function?

Scripts/models/logit.py
```diff
@@ -413,6 +413,8 @@ def calc_individual_mode_prob(self, is_car_user, zone):
         """
```
zptro, line 413: Docstring should be updated.

zptro (changes requested):
I am impressed by your ability to dive deep into the core of the demand model and add meaningful code! A very good PR.

However, some things could be further developed.

Scripts/datatypes/zone.py
```diff
@@ -0,0 +1,27 @@
...
+        for area in areas:
+            if isinstance(areas[area][0], tuple):
+                for interval in areas[area]:
+                    if interval[0] <= number < interval[1]:
+                        self.area = area
+                        break
+            else:
+                if areas[area][0] <= number < areas[area][1]:
+                    self.area = area
+                    break
```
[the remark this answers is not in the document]
atte-wsp: Nice! I made sort of a quick fix here, good that there already was better solution available. 

However, I was thinking that should area intervals be mutually exclusive? If I'm correct looping trough dictionary keys might end up returning either surrounding or surround_other / surround_train

Scripts/demand/trips.py
```diff
@@ -90,13 +89,14 @@ def create_population(self):
...
         """
+        numpy.random.seed(param.population_draw)
         bounds = slice(0, self.zone_data.first_peripheral_zone)
         self.incmod = linear.IncomeModel(
             self.zone_data, bounds, self.age_groups, self.resultdata)
         self.population = []
         zones = self.zone_data.zone_numbers[bounds]
         self.zone_population = pandas.Series(0, zones)
-        for zone_number in zones:
+        for zone_number in self.zone_data.zone_numbers[bounds]:
```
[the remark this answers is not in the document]
atte-wsp: Yep, sorry for these simple mistakes. Some unintentional changes due to replacing old branch.

Scripts/modelsystem.py
```diff
@@ -525,11 +528,37 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
...
+                    zone = person.zone
+                    result_dict[attr].append(getattr(zone, attr))
+                else:
+                    result_dict[attr].append(getattr(person, attr))
+        results = pandas.DataFrame(result_dict)
+        for attr in results.columns:
+            self.resultdata.print_data(
+                pandas.Series(results[attr]), "agents.txt", attr)
+        random.seed(None)   
+        log.info("Results printed to file agents.txt")
```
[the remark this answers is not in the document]
atte-wsp: That sounds more elegant solution to collect results from objects. Using dict and data frame creates unnecessary temporary variables, if one could use object methods.

Scripts/datatypes/zone.py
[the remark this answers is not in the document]
zptro: Good point. Maybe we should loop over `parameters.zone.area_aggregation`?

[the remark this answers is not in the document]
atte-wsp: Yep, that's a good and clear solution.

atte-wsp:
Changes proposed in review are done. Some problems exist with accessibility definitions, but I'd think it is better to fix them with separate PRs.

### Marginal utility of money 
Accessibility measure does not consider difference in marginal utility of money across income groups. Especially this is highlighted with congestion charges. This would probably need re-scaling of logsum values, but there is no simple solution for this.

Theoretically, could it be beneficial to use variable `cost/income` ([Train and McFadden 1978](https://eml.berkeley.edu/reprints/mcfadden/goods_leisure.pdf)) as part of utility functions in place of cost? 

### Level effect in accessibility
Another problem is that the logsum accessibility of tours does not have a consistent benchmark utility for separately estimated models. This means that accessibility levels for different travel purposes and areas are not comparable. 

When comparing utilities between scenarios the benchmark utility of each individual cancels out while computing the difference. However, this level effect restricts comparing `accessibility gap` and `minimum accessibility` between income groups, as for example age groups make different amount of tours with certain travel purpose (for 30-44 yo person there is higher proportion of work related trips and 7-18 yo would have high proportion of school trips).

I'd think there are three options to overcome this:
1) Descale accessibility to common benchmark using method proposed in Dong et al. 2006 and Miller 2020. 
2) Make comparison with only one tour purpose (income groups / home-work tours, etc).
3) Abandon logsum measure and use generalized time differences of trips made by agent.

Scripts/models/logit.py
```diff
@@ -426,10 +441,32 @@ def calc_individual_mode_prob(self, is_car_user, zone):
...
+                car_expsum += mode_exps[mode]
+            else:
+                sust_expsum += mode_exps[mode]
         probs = numpy.empty(len(modes))
         for i, mode in enumerate(modes):
             probs[i] = mode_exps[mode] / mode_expsum
-        return probs
+        # utils to money
+        logsum = numpy.log(mode_expsum)
+        sust_logsum = numpy.log(sust_expsum)
```
zptro, line 453: These are expensive calculations, and I might consider some optimization even though readability would suffer. What is the run time right now?

In the current model, `sust_expsum` is the same for all agents (because car usership only modifies `car_logsum`), so we could get this from `purpose.sustainable_accessibility`.

Scripts/models/logit.py
```diff
@@ -426,10 +441,32 @@ def calc_individual_mode_prob(self, is_car_user, zone):
...
+        car_logsum = numpy.log(car_expsum)
+        b = self._get_cost_util_coefficient()
+        if isinstance(b, tuple):
```
zptro, line 456: I have used try-except clauses for these, as they are faster for the standard case, and might be considered more pythonical (https://docs.python.org/3.5/glossary.html#term-eafp).

Scripts/demand/trips.py
```diff
@@ -90,12 +89,14 @@ def create_population(self):
         list
             Person
         """
+        numpy.random.seed(param.population_draw)
```
zptro, line 92: The other random seed is set in `ModelSystem`. Maybe we could put them in the same place?

Scripts/datatypes/person.py
```diff
@@ -119,3 +120,31 @@ def add_tours(self, purposes, tour_probs):
                 if random.random() < non_home_prob:
                     non_home_tour = Tour(purposes["oo"], tour)
                     self.tours.append(non_home_tour)
+
+    def write_file(self, resultdata, filename):
```
zptro, line 124: I find the passing of `resultdata` and `filename` to this function a bit unnecessary. Instead we could return the line as a `str` and run `resultdata.print_line()` in `ModelSystem`. 

And a very pythonic name for the function would then be `__str__()`.

Scripts/modelsystem.py
```diff
@@ -538,10 +541,21 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
...
+                "total_access", "sustainable_access", "car_access"
+                ])
+            fname = "agents"
+            self.resultdata.print_line(line, fname)
             for person in self.dm.population:
                 person.calc_income()
-        log.info("Demand calculation completed")
+                person.write_file(self.resultdata, fname)
+            log.info("Results printed to file ".format(fname)) 
+        log.info("Demand calculation completed")
```
zptro, line 558: Avoid adding extra white space.

Scripts/modelsystem.py
```diff
@@ -538,10 +541,21 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
...
                 self.dtm.add_demand(tour)
         if is_last_iteration:
+            random.seed(zone_param.population_draw)
             self.dm.incmod.predict()
+            random.seed(None) 
+            line = "\t".join([
+                "person_id", "age_group", "gender", "car_user", "income",
+                "number", "area", "municipality", "nr_tours",
+                "total_access", "sustainable_access", "car_access"
+                ])
```
zptro, line 551: Maybe move this to `Person` as class variable?

Scripts/datatypes/person.py
```diff
@@ -119,3 +120,31 @@ def add_tours(self, purposes, tour_probs):
...
+            total_access += tour.total_accessibility
+            sustainable_access += tour.sustainable_accessibility
+            car_access += tour.car_accessibility
+        # print to file
+        persondata = "{:d}\t{:s}\t{:s}\t{!s}\t{:.0f}".format(
+            self.id, self.age_group, self.gender, self.is_car_user, self.income)
+        zonedata = "{:d}\t{:s}\t{:s}".format(
+            self.zone.number, self.zone.area, self.zone.municipality)
+        accessdata = "{:d}\t{:.1f}\t{:.1f}\t{:.1f}".format(
+            nr_tours, total_access, sustainable_access, car_access)
```
zptro, line 148: `nr_tours` could be replaced by `len(tours)`.

zptro (changes requested):
Overall this looks very good now. Theoretical issues can be handled in other PR as you suggested.

Some minor comments:

Scripts/demand/trips.py
atte-wsp: This needs to be set here because `numpy` module (unlike `random`) inits random seed when module is imported. 

Moving to Python 3 helps as then we can use `random.choices` for weighted random choices.

Scripts/modelsystem.py
atte-wsp: Good idea. 

I wasn't happy having two list definitions even in same class, so decided to refactor fixed lists from `Person.__str__() `.

Scripts/models/logit.py
atte-wsp: Assigning mode and destination for agents is around ~ 8 mins.

Getting sustainable accessibility from `purpose` works nicely and I think it is easy to follow. Also, `car_logsum ` is no longer used so it can be removed. However, this does not have much effect on running time. 

I think better efficiency would require refactoring whole `calc_individual_mode_prob` to use pre-calculated utility and logsum vectors for car users and others.

Scripts/modelsystem.py
```diff
@@ -543,19 +543,15 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
...
             fname = "agents"
-            self.resultdata.print_line(line, fname)
+            header = "\t".join(self.dm.population[0].attr)
+            self.resultdata.print_line(header, fname)
             for person in self.dm.population:
                 person.calc_income()
-                person.write_file(self.resultdata, fname)
-            log.info("Results printed to file ".format(fname)) 
-        log.info("Demand calculation completed")        
+                self.resultdata.print_line(person.__str__(), fname)
```
zptro, line 552: The `__str__()` function is usually called like this: `str(person)`.

Scripts/datatypes/person.py
```diff
@@ -120,31 +126,23 @@ def add_tours(self, purposes, tour_probs):
...
-            self.id, self.age_group, self.gender, self.is_car_user, self.income)
-        zonedata = "{:d}\t{:s}\t{:s}".format(
-            self.zone.number, self.zone.area, self.zone.municipality)
-        accessdata = "{:d}\t{:.1f}\t{:.1f}\t{:.1f}".format(
-            nr_tours, total_access, sustainable_access, car_access)
-        resultdata.print_line("\t".join(
-            [persondata, zonedata, accessdata]), filename)
+        persondata = [str(getattr(self, attr)) for attr in Person.person_attr]
+        zonedata = [str(getattr(self.zone, attr)) for attr in Person.zone_attr]
+        tourdata = [str(tour_attributes[attr]) for attr in Person.tour_attr]
```
zptro, line 147: Very nice solution!

zptro (approved):

atte-wsp:
@zptro Added last style updates and this PR is ready

zptro (approved):

PR #277, opened 2021-03-29
Fix filename

johpiip:
Previously, transit distances were overwritten by transit times in EmmeAssignment.

@zptro, is "minutes" the correct unit?

Not tested with Emme!


zptro (commented):
Transit distance is not overwritten, because transit time is stored in a second column named "time". I agree the file name "transit_kms.txt" is confusing, but maybe they could still be in the same file?

johpiip:
@zptro Ah sorry, I misinterpreted what `print_data` does because I have different output data habits. I will remove this PR! :)

PR #278, opened 2021-03-30
Integrate batin functionality

zptro:
- Integrate volume-delay function and transit time function settings into assignment
- Remove modification and usage of attributes `hdw` and `@hinta` and use time-period specific extra attributes directly


johpiip (approved):
Looks good to me! Great documentation in `Scripts/parameters/assignment.py`!

PR #279, opened 2021-03-31
Assignment log messages

atte-wsp:
We've had issues with assignment convergence and consequently thought that log messages in UI would be helpful for many users. Also, removing several create attribute messages so that Emme Logbook is easier to use.


zptro (approved):
Very nice! Should be part of official release.

PR #282, opened 2021-04-07
Print line feature

zptro:
Can be used for printing a descriptive row of text for each agent.

Also adds better formatting when printing dataframes.


zptro:
Moved column alignment to own PR: #283

atte-wsp (approved):
Enables more efficient and easy way of writing agent data to file. Good work.

Only thing that I was thinking if there is any good solution for writing column titles? On the other hand it is easy to write title row before looping over agents.

PR #283, opened 2021-04-08
Align columns better with values padded up to 12 digits

zptro:
- Align columns better with values padded up to 12 digits
- Restrict number of significant digits to 7, larger numbers as exponents

### DIFF
--- a/Scripts/datahandling/resultdata.py
+++ b/Scripts/datahandling/resultdata.py
@@ -27,9 +27,13 @@ class ResultsData:
                     f.write(row)
         self._list_buffer = {}
         for filename in self._df_buffer:
-            self._df_buffer[filename].to_csv(
+            df = self._df_buffer[filename]
+            df.index = df.index.astype(str, copy=False)
+            # Pad index according to its longest item
+            df.index = df.index.str.pad(len(max(df.index, key=len)))
+            df.to_csv(
                 os.path.join(self.path, filename),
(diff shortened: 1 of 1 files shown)


johpiip (commented):
When decimal numbers are rounded to 7 significant digits, trailing zeros are dropped. This can be verified in `workforce_accessibility_per_area.txt` where `helsinki_cbd	4571.64987`becomes `espoo_vant_kau	    3993.185` (7 significant digits = 7 digits) but `helsinki_cbd	4571.64987` becomes `  helsinki_cbd	     4571.65` (7 significant digits = 6 digits because last one is rounded to zero). Is this intentional? Both ways are fine by me.

zptro:
Should ask users if they want this change or not.

johpiip:
@zptro My comment in Teams was about whitespaces but [my comment in this PR](https://github.com/HSLdevcom/helmet-model-system/pull/283#pullrequestreview-631144186) was about significant digits. I like the idea of explicitly rounding to seven significant digits but I commented on trailing zeros. :)

zptro:
> @zptro My comment in Teams was about whitespaces but [my comment in this PR](https://github.com/HSLdevcom/helmet-model-system/pull/283#pullrequestreview-631144186) was about significant digits. I like the idea of explicitly rounding to seven significant digits but I commented on trailing zeros. :)

Yes, maybe I could do a separate PR, changing `%1.5f` into `%.7g`.

PR #284, opened 2021-04-08
Add vans and save demand matrices before last assignment

zptro:
This is hopefully the last time I change this. The idea is to save all demand matrices before end assignment starts, so that end assignment can be re-run if it crashes for some reason. The first time I implemented this (https://github.com/HSLdevcom/helmet-model-system/pull/237/commits/05c91e538bbb0f0bbd380cb2fc6b088dbdd6bdd9), I had forgot that vans are added later. When I tried to fix this problem, I just moved saving of demand matrices back into the assignment loop, so the original idea was lost.

### DIFF
--- a/Scripts/modelsystem.py
+++ b/Scripts/modelsystem.py
@@ -291,14 +291,15 @@ class ModelSystem:
                 trip_sum[mode] / sum_all, "origins_shares.txt", mode)
             mode_shares[mode] = trip_sum[mode].sum() / sum_all.sum()
         self.mode_share.append(mode_shares)
+        for ap in self.ass_model.assignment_periods:
+            self.dtm.add_vans(ap.name, self.zdata_forecast.nr_zones)
+            if iteration=="last":
+                self._save_demand_to_omx(ap.name)
 
         # Calculate and return traffic impedance
         for ap in self.ass_model.assignment_periods:
             tp = ap.name
             log.info("Assigning period " + tp)
-            self.dtm.add_vans(tp, self.zdata_forecast.nr_zones)
-            if iteration=="last":
-                self._save_demand_to_omx(ap.name)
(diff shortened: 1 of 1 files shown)


Scripts/modelsystem.py
```diff
@@ -292,13 +292,16 @@ def run_iteration(self, previous_iter_impedance, iteration=None):
...
+        # Add vans and save demand matrices
+        for ap in self.ass_model.assignment_periods:
+            self.dtm.add_vans(ap.name, self.zdata_forecast.nr_zones)
+            if iteration=="last":
+                self._save_demand_to_omx(ap.name)
+
         # Calculate and return traffic impedance
         for ap in self.ass_model.assignment_periods:
             tp = ap.name
             log.info("Assigning period " + tp)
```
johpiip, line 304: I am not sure if I understand correctly but it looks to me that these two lines are the only ones that can fail between `master` and this PR. So is this just a refactoring (= saving all demand matrices before printing "Assigning period xxx") and not really bugfixing?

johpiip (commented):
Not sure if I understand correctly but it looks to me that the only lines that can fail between `master` and this branch are `tp = ap.name` and ``

Scripts/modelsystem.py
zptro: Well, this can fail (badly): https://github.com/HSLdevcom/helmet-model-system/blob/76fdd8c158a1dc543b343250ca5bc3d541476c39/Scripts/modelsystem.py#L305

And because it runs in a loop, having demand matrix saving in the same loop means it will then not be run for consecutive iterations.

johpiip: Ah, I see! If line 305 fails during "aht", there would not be demand matrices saved for "iht" and "pt" - but with this PR, there would be. Good work!

johpiip (approved):

PR #286, opened 2021-04-09
Save transit congestion effect

atte-wsp:
Save transit volumes before congested assignment on links (`@transit_uncongested`). Attribute can be used to visualize the effect of crowding on route choices.

Benefits:
* Planners are often interested of which part of network would be congested if there was no route changes due to crowding (more network capacity could add benefits).
* Congested assignment is new feature, therefore it is beneficial to understands its effects on assignment results.

New extra attribute is required. However, this information cannot be retrieved in post congested assignment analyses.


atte-wsp:
@samulikyytsonen

zptro (dismissed):
This is probably a good addition, but might be regarded as more than a patch, as it may cause existing model setups to crash (if more extra attributes are not added). Hence I schedule it for the Helmet 4.1 release.

zptro:
One alternative would be to have an option to run end assignment without congestion. In Helmet 4.1 copying scenarios will be much easier and the user could then just copy the network to a new emme scenario and re-run end assignment without congestion.

zptro:
Replaced by #313.

PR #289, opened 2021-04-13
Create test environment for emme assignment

zptro:
Created mock-up version of `EmmeProject`. Emulates key features of EMME API, so that `EmmeAssignmentModel` and `AssignmentPeriod` can be unit tested without EMME binding.

### DIFF
--- a/Scripts/assignment/emme_bindings/mock_project.py
+++ b/Scripts/assignment/emme_bindings/mock_project.py
@@ -37,8 +37,12 @@ class MockProject:
                 scenario.extra_attribute(extra_attribute_name).initialize(
                     extra_attribute_default_value)
 
-    def copy_matrix(self, *args, **kwargs):
-        pass
+    def copy_matrix(self, from_matrix, matrix_id, matrix_name,
+                    matrix_description):
+        self.create_matrix(matrix_id, matrix_name, matrix_description)
+        eb = self.modeller.emmebank
+        eb.matrix(matrix_id).set_numpy_data(
+            eb.matrix(from_matrix).get_numpy_data())
 
     def network_calc(self, *args, **kwargs):
         pass
--- a/Scripts/assignment/emme_bindings/mock_project.py
+++ b/Scripts/assignment/emme_bindings/mock_project.py
@@ -1,0 +1,125 @@
+from collections import namedtuple
+
+
+class MockProject:
+    def __init__(self):
+        Modeller = namedtuple("Modeller", "emmebank")
+        self.modeller = Modeller(EmmeBank())
+
+    def create_matrix(*args, **kwargs):
+        pass
+
+    def create_extra_attribute(*args, **kwargs):
+        pass
+
+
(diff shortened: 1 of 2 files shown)


Scripts/assignment/emme_bindings/mock_project.py
```diff
@@ -0,0 +1,423 @@
...
+
+
+class MockProject:
+    """Mock-up version of `EmmeProject`.
+
+    Emulates key features of EMME API, so that `EmmeAssignmentModel`
+    and `AssignmentPeriod` can be unit tested without EMME binding.
+    """
+
+    path = "c:\\xxx\\yyy"
```
johpiip, line 12: What is the meaning of this `path`?

johpiip (approved):
Looks good to me! VSCode finds new tests and they give no errors. Just checking: is all the code in `mock_project.py` written by us and not copied from the real classes?

I came across "old-style and new-style classes" when googling for differences between Python 2 and 3. [Here](https://portingguide.readthedocs.io/en/latest/classes.html) is an example. Do you think this will be a problem later?

Edit. Old-style: `class OldStyle: pass` and new-style: `class NewStyle(object)`.

johpiip:
Additional review comment, I only notices this now! Results are printed to `tests/test_data/Results/2016_test` instead of `tests/test_data/Results/test` where all other results go. That folder is not gitignored so they clutter git status.

zptro:
> Looks good to me! VSCode finds new tests and they give no errors. Just checking: is all the code in `mock_project.py` written by us and not copied from the real classes?

Yes, all written by us based on the Emme documentation. The real classes are (I think) compiled and cannot be inspected.
 
> I came across "old-style and new-style classes" when googling for differences between Python 2 and 3. [Here](https://portingguide.readthedocs.io/en/latest/classes.html) is an example. Do you think this will be a problem later?

These classes do not inherit anything from `object`. There are only a few such classes in Helmet.

> Additional review comment, I only notices this now! Results are printed to tests/test_data/Results/2016_test instead of tests/test_data/Results/test where all other results go. That folder is not gitignored so they clutter git status.

Good point! I will fix this!

Scripts/assignment/emme_bindings/mock_project.py
zptro: In `EmmeProject`, it refers to the directory path of the *emmebank* file and it is used to locate the function files here: https://github.com/HSLdevcom/helmet-model-system/blob/8d6073468da8c7519fd575edbd809e5804219707/Scripts/assignment/assignment_period.py#L460

I will remove this dependency in the Helmet 4.1 development.

johpiip:
> These classes do not inherit anything from `object`. There are only a few such classes in Helmet.

But they will inherit in Python 3, implicitly, based on the docs. Well, this is just one thing to look at when moving from Python 2 to 3. 😄

johpiip (approved):
We discussed this on Monday. Looks good to me!

PR #290, opened 2021-04-14
Refactor extra attribute creation

zptro:
- Remove hard-coded extra attribute creation
   - Automatically create link attributes for all demand classes and for transit segment results
- Automatically add time-period tag to extra attributes (comes in handy if same emme scenario is used for all time periods)
- Move bike assignment from own emme scenario to the scenario that other modes uses
   - Extra attributes for bike consistent with other modes
   - Requires modes to be modified, because only one mode (`mode.type = "AUTO"`) in each scenario can be used for primary assignment: One synthetic AUTO mode `h` that is assigned to either car or bike.
- Update test network
  - New mode `h`
  - New network transaction format
- Remove obsolete macros


johpiip (approved):
I trust you know what you are doing with this large PR so I am approving. 😄 Can you explain here or on Teams what `Database/Scenario_19` files mean?

zptro:
I will move the test files in `database/scenario_19` to test folder in the next PR. They can be read with Modeller to form a test network in Emme (and in the future in `MockProject` too).

johpiip (approved):

PR #308, opened 2021-05-10
Agent tours to separate file

atte-wsp:
Option to write agent data using separate file for tours. Separate person and trip tables is common way of storing travel survey data, so combining them in R is simple task.

+ More flexible analysis possibilities in R. Therefore no need to change Helmet model system, if mode specific extra analysis is needed (for example mode shares by gender).
- Consumes more hard disk space (from ~100 MB to 200 MB).

### DIFF
--- a/Scripts/modelsystem.py
+++ b/Scripts/modelsystem.py
@@ -492,7 +492,7 @@ class AgentModelSystem(ModelSystem):
             0, self.zdata_forecast.zone_numbers[self.dm.cm.bounds])
         for person in self.dm.population:
             person.decide_car_use()
-            car_users[person.zone] += person.is_car_user
+            car_users[person.zone.number] += person.is_car_user
             person.add_tours(self.dm.purpose_dict, tour_probs)
             for tour in person.tours:
                 tour.choose_mode(person.is_car_user)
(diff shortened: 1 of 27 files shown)


atte-wsp:
This update is not necessary in our distribution analysis, but could be better in long term?

Scripts/modelsystem.py
```diff
@@ -544,13 +544,23 @@ def _add_internal_demand(self, previous_iter_impedance, is_last_iteration):
...
+            fname0 = "agents"
+            fname1 = "tours"
+            # loop to first person that has tours
+            for person0 in self.dm.population:
+                if isinstance(person0.tours, list):
+                    break
+            # print person and tour attr to files
+            self.resultdata.print_line("\t".join(person0.attr), fname0)
+            self.resultdata.print_line(
+                "\t".join(person0.tours[0].attr), fname1)
```
zptro, line 556: We can get the attribute names directly from `Person` and `Tour`:

````
from datatypes.person import Person
from datatypes.tour import Tour


self.resultdata.print_line("\t".join(Person.attr), fname0)
self.resultdata.print_line("\t".join(Tour.attr), fname1)
````

The loop is hence unnecessary (and probably not even working as intended, `person0.tours` is always a `list`, whether it is empty or not, the pythonical way to check if a `list` is non-empty is simply `if person0.tours`).

zptro (changes requested):
I guess this change is alright. Just one small comment.

Scripts/modelsystem.py
atte-wsp: Thanks! I'm glad that there is a better solution as the one I implemented previously was quite complex way of getting headers for output files.

zptro (approved):

PR #309, opened 2021-05-11
Split income model in two: Helsinki and other

zptro:
Helsinki model requires new Lauttasaari dummy and storing of mode choice logsum in `ZoneData`.

### DIFF
--- a/Scripts/parameters/income.py
+++ b/Scripts/parameters/income.py
@@ -23,7 +23,7 @@ log_income = {
     "standard_deviation": 0.35499,
 }
 log_income_helsinki = {
-    "constant": 0,
+    "constant": 7.50592,
     "zone": {
         "cbd": 0.26469,
         "lauttasaari": 0.25522,
--- a/Scripts/parameters/income.py
+++ b/Scripts/parameters/income.py
@@ -1,24 +1,39 @@
 log_income = {
-    "constant": 6.52186,
+    "constant": 6.8611,
     "zone": {
-        "perc_detached_houses_sqrt": 0.16309,
-        "hw_t": 0.07543,
-        "cbd": 0.13188,
+        "perc_detached_houses_sqrt": 0.18717,
+        "hw": 0.04091,
     },
-    "car_users": 0.21289,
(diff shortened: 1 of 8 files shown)


johpiip (approved):

atte-wsp (changes requested):
Test run showed that there is large difference in incomes for Helsinki and other region. For Helsinki result incomes are between 0-5 and for others 0-20000. 

Should there be constant in Helsinki parameters?

Scripts/parameters/income.py
```diff
@@ -1,24 +1,39 @@
...
+        "Hyvinkaa": -0.08642,
+        "Jarvenpaa": -0.13453,
+        "Sipoo": 0.12064,
+        "Tuusula": -0.09369,
+        "Vihti": -0.07904,
     },
+    "standard_deviation": 0.35499,
+}
+log_income_helsinki = {
+    "constant": 0,
```
atte-wsp, line 26: Is this correct? I think this causes large difference between Helsinki / other.

zptro: Great that you found this error! The constant had indeed disappeared from Lauri's excel file.

atte-wsp (approved):
Model predictions are ok. Median income is still constantly lower in Helsinki than rest of the region (slides in SharePoint), but I'm not sure if this is error or based on statistics.

Scripts/parameters/income.py
```diff
@@ -1,24 +1,39 @@
 log_income = {
-    "constant": 6.52186,
+    "constant": 6.8611,
     "zone": {
-        "perc_detached_houses_sqrt": 0.16309,
-        "hw_t": 0.07543,
-        "cbd": 0.13188,
+        "perc_detached_houses_sqrt": 0.18717,
```
atte-wsp, line 4: Check scale as Zonedata defines share as `self["perc_detached_houses_sqrt"] = (100*landdata["detach"]) ** 0.5`

atte-wsp (changes requested):

atte-wsp (approved):

PR #310, opened 2021-05-11
Fix bug related to python2to3 transition

zptro:


johpiip (approved):
Nice catch! Was this that bug that happened yesterday in the demo?

zptro:
> Nice catch! Was this that bug that happened yesterday in the demo?

Yup, this was it.

PR #311, opened 2021-05-12
Separate accessibility calculation class

zptro:
Refactor accessibility calculations and printing to new sub-class `AccessibilityModel(ModeDestModel)`. The benefits are:
- Accessibility calculations do not clutter demand model calculations
- Separate parameters for capital region and surrounding municipalities can be ignored in accessibility calculations

Also add printing of all-purposes-combined sustainable and car accessibility.

### DIFF
--- a/Scripts/datatypes/purpose.py
+++ b/Scripts/datatypes/purpose.py
@@ -121,7 +121,7 @@ class TourPurpose(Purpose):
         else:
             self.model = logit.ModeDestModel(
                 zone_data, self, resultdata, is_agent_model)
-            self.access_model = logit.AccessibilityModel(
+            self.accessibility_model = logit.AccessibilityModel(
                 zone_data, self, resultdata, is_agent_model)
         self.modes = self.model.mode_choice_param.keys()
         self.histograms = {mode: TourLengthHistogram() for mode in self.modes}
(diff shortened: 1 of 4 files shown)


Scripts/datatypes/purpose.py
```diff
@@ -121,6 +121,8 @@ def __init__(self, specification, zone_data, resultdata, is_agent_model):
         else:
             self.model = logit.ModeDestModel(
                 zone_data, self, resultdata, is_agent_model)
+            self.access_model = logit.AccessibilityModel(
```
johpiip, line 124: Consider renaming the variable as `accessibility_model` instead of `access_model`.

johpiip (approved):
Nice work! I looked at the test results and noticed that files that loop through tour purposes always print `wh` class and it might cause misunderstandings. Examples would be `attraction.txt` and `trip_lengths.txt`. This behavior was there already before this PR.

zptro:
> Nice work! I looked at the test results and noticed that files that loop through tour purposes always print `wh` class and it might cause misunderstandings. Examples would be `attraction.txt` and `trip_lengths.txt`. This behavior was there already before this PR.

Yes, you are right. I think this line is unneccessary to run for `wh`:
https://github.com/HSLdevcom/helmet-model-system/blob/f8cad5a782bfb39f4bdb6a6ed5559e26c16f2c6f/Scripts/modelsystem.py#L135

johpiip (approved):
Great, `wh` purposes are now gone from results. :)

PR #318, opened 2021-06-01
Read test network from files

zptro:
Read test network used in `MockProject` from test files used also for `EmmeAssignment` tests with EMME. Enables more versatile tests of network handling and assignment. The last commit is an excellent proof-of-concept, where a bug in link length aggregation was previously not detected because the test network was too limited in features.


johpiip (approved):
Great!

PR #323, opened 2021-06-07
Calculate tour costs

atte-wsp:
* Generalized and out-of-pocket costs of tours based on choices

### DIFF
--- a/Scripts/datatypes/tour.py
+++ b/Scripts/datatypes/tour.py
@@ -4,7 +4,7 @@ import random
 import parameters.car as param
 import parameters.zone as zone_param
 from parameters.assignment import assignment_classes, vot_inv
-from parameters.impedance_transformation import divided_classes
+from parameters.impedance_transformation import divided_classes, trips_month
 
 
 class Tour(object):
@@ -199,7 +199,7 @@ class Tour(object):
         self.gen_cost = self.cost + time * vot
 
     def _get_cost(self, impedance, mtx_type):
-        """Check if matrix and return value from position. """
+        """Get cost and time components from tour dest choice."""
         if self.mode in divided_classes:
             ass_class = "{}_{}".format(
                 self.mode, assignment_classes[self.purpose.name])
@@ -215,7 +215,7 @@ class Tour(object):
                 departure_imp = impedance["pt"][mtx_type][ass_class]
                 sec_dest_imp = impedance["pt"][mtx_type][ass_class]
                 return_imp = impedance["pt"][mtx_type][ass_class]
-            # first leg
+            # first leg of tour
             cost += departure_imp[self.position[0], self.position[1]]
             # check if tour has secondary destination and add accordingly
             if len(self.position) > 2:
@@ -225,6 +225,10 @@ class Tour(object):
                 cost += return_imp[self.position[1], self.position[0]]
         except KeyError:
             pass
+        # scale transit costs from month to day
+        if self.mode == "transit" and mtx_type == "cost":
+            idx = int(self.position[0] > self.purpose.zone_data.first_surrounding_zone)
+            cost /= trips_month[ass_class][idx]
         return cost
 
     def __str__(self):
--- a/Scripts/datatypes/tour.py
+++ b/Scripts/datatypes/tour.py
@@ -224,6 +224,8 @@ class Tour(object):
             else: 
                 cost += return_imp[self.position[1], self.position[0]]
         except KeyError:
+            # bike and walk modes do not have cost matrices specified
+            # KeyErrors are produced when trying to access matrix
             pass
         # scale transit costs from month to day
         if self.mode == "transit" and mtx_type == "cost":
--- a/Scripts/datatypes/tour.py
+++ b/Scripts/datatypes/tour.py
@@ -3,7 +3,7 @@ import random
 
 import parameters.car as param
 import parameters.zone as zone_param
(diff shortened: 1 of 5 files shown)


Scripts/datatypes/tour.py
```diff
@@ -179,6 +181,24 @@ def choose_secondary_destination(self, cumulative_probs):
                     + numpy.searchsorted(cumulative_probs, self._sec_dest_draw))
         self.position = (self.position[0], self.position[1], dest_idx)
         self.purpose.sec_dest_purpose.attracted_tours[self.mode][dest_idx] += 1
+    
+    def calc_cost(self):
+        """Construct cost and time components from tour dest choice. """
+        impedance = self.purpose.model.impedance[self.mode]
+        time = self._get_cost(impedance["time"])
+        self.cost = self._get_cost(impedance["cost"])
```
zptro, line 189: Will this not cause a `KeyError` for bike and walk?

zptro (commented):
Nice attempt! The secondary destinations are not handled exactly right (see comment below) and memory consumption will be high. I am thinking, should we have a simpler approach using the un-transposed impedance matrices (e.g., assuming aht-iht pattern for all work tours) and adding tour legs one-by-one?

Scripts/datatypes/tour.py
```diff
@@ -179,6 +181,24 @@ def choose_secondary_destination(self, cumulative_probs):
...
+        self.gen_cost = self.cost + time * vot[self.purpose.name]
+
+    def _get_cost(self, mtx):
+        """Check if matrix and return value from position. """
+        cost = 0
+        if isinstance(mtx, numpy.ndarray):
```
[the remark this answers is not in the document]
atte-wsp: For walk and bike `mtx` is sometimes float zero and there would be index error. Probably this should be handled with try-except clause?

Scripts/datatypes/tour.py
atte-wsp: I think this is related to above. Bike and walk return float zeros.

Scripts/parameters/impedance_transformation.py
```diff
@@ -275,3 +275,17 @@
...
+    "hw": 0.132, # eur / min
+    "hu": 0.132,
+    "hc": 0.132,
+    "hs": 0.0895,
+    "ho": 0.0895,
+    "wo": 0.0895,
+    "woo": 0.0895,
+    "oo": 0.0895,
+    "wh": 0.0895,
+}
```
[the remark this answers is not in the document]
atte-wsp: True, I made this as a quick first draft and forgot to fix it later.

Scripts/models/logit.py
```diff
@@ -318,6 +318,7 @@ def calc_basic_prob(self, impedance):
             Mode (car/transit/bike/walk) : numpy 2-d matrix
                 Choice probabilities
         """
+        self.impedance = impedance
```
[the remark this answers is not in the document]
atte-wsp: True, after monitoring the memory consumption this does need some changes.  Didn't realize that impact on memory consumption is that big. Probably refactoring with simpler solution is the best option as secondary destination require changes as well.

atte-wsp:
@zptro Made new version with simplified approach. Would this be more efficient? Not sure if I'm completely happy with this implementation as the cost calculation has quite many exceptions, but could not figure out better solutions at this point.

Results are reasonable and highlight more differences between Helsinki cbd and other region than the previous implementation.

Scripts/datatypes/tour.py
```diff
@@ -179,6 +182,54 @@ def choose_secondary_destination(self, cumulative_probs):
...
+            # first leg of tour
+            cost += departure_imp[self.position[0], self.position[1]]
+            # check if tour has secondary destination and add accordingly
+            if len(self.position) > 2:
+                cost += sec_dest_imp[self.position[1], self.position[2]]
+                cost += return_imp[self.position[2], self.position[0]]
+            else: 
+                cost += return_imp[self.position[1], self.position[0]]
+        except KeyError:
+            pass
```
zptro, line 227: I think this try-except clause could be considered to wide, we risk excepting unintended bugs. But I cannot come up with a more elegant solution.

Scripts/transform/impedance_transformer.py
```diff
@@ -47,11 +47,9 @@ def transform(self, purpose, impedance):
...
-                day_imp["transit"]["cost"], trips_month["transit_work"][0])
-            trips_per_month[purpose.ubounds, :] = trips_month["transit_work"][1]
-            day_imp["transit"]["cost"] /= trips_per_month
-        else:
-            day_imp["transit"]["cost"] /= trips_month["transit_leisure"]
+        transit_class = "{}_{}".format("transit", assignment_classes[purpose.name])
+        trips_per_month = numpy.full_like(
+            day_imp["transit"]["cost"], trips_month[transit_class][0])
+        trips_per_month[purpose.ubounds, :] = trips_month[transit_class][1]
+        day_imp["transit"]["cost"] /= trips_per_month
```
zptro, line 54: This generalization is nice!

zptro (approved):
Yes, I think this should be a working and stable solution. Well done!

Scripts/datatypes/tour.py
zptro: Maybe we could add a comment describing what the except clause is supposed to catch?

PR #334, opened 2021-06-16
Get AssignmentPeriod dist unit cost from zonedata

### LINKED ISSUE #333: Distance unit cost not set from zone input data
Distance unit cost should be set via zone input data and it is used by `AssignmentPeriod` class. Currently model system only changes `dist_unit_cost` for `EmmeAssignmentModel` and `AssignmentPeriod` cost is fixed value from parameters.

atte-wsp:
* Distance unit cost currently used in AssignmentPeriod (not EmmeAssignment)

Closes #333


zptro (approved):
Great! Thanks for fixing this!

PR #336, opened 2021-06-17
Transit fare class

zptro:
Instead of passing a dictionary of transit fares and fare exceptions around, this PR introduces a separate class for transit fares.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -171,14 +171,8 @@ class AssignmentPeriod(Period):
         
         Parameters
         ----------
-        fares : dict
-            key : str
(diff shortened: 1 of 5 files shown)


johpiip (approved):
Nice improvement! Just one thing: `.tco` file has a term "transit zone monthly **cost**" but this class is "transit **fare**". Should we use only one of them?

zptro:
> Nice improvement! Just one thing: `.tco` file has a term "transit zone monthly **cost**" but this class is "transit **fare**". Should we use only one of them?

Yes, you are right, it would be better to talk about "transit fare" in the input file description. However, input file metadata is not really part of model development and could actually be in Finnish just as well.

PR #344, opened 2021-07-09
Fix TypeError when running the model in new Helmet-UI using MockAssignment

johpiip:
This PR fixes an issue where running Helmet 4.1.0-alpha.2 in new Helmet-UI (https://github.com/HSLdevcom/helmet-ui/pull/100) using MockAssignment would fail. Lines that are fixed in this PR raised the following exception:
```
[ERROR] TypeError: can only concatenate str (not "bytes") to str
```

This error does not appear when running the model in Helmet-UI v4.0.7 Alppila Edition (no matter the model version), nor when running Helmet 4.0.6 in new Helmet-UI.

I am marking this as draft as I need to make sure if updating `python-shell` npm package in Helmet-UI breaks something different.


johpiip:
Upgrading `python-shell` makes bad stuff happen (https://github.com/HSLdevcom/helmet-ui/issues/101) but the model run still succeeds. I will mark this as "ready for review" and I will fix new issues in another PR.

zptro (approved):

PR #349, opened 2021-08-16
Add car ownership sensitivity analyses

zptro:
Expands `helmet.py` into three different model runs (baseline, car++ and car--), which all produce separate result files

- car++ increases car density by 0.1 (100 vehicles per 1000 inhabitants) everywhere
- car-- decreases car density by 20 %


johpiip (approved):

johpiip (approved):
Quick fix! :)

PR #352, opened 2021-08-17
Car assignment last in end assigment

zptro:
Do bike assignment before car assignment in last iteration so that car results are stored in `timau` and `volau` (when results are stored in different EMME scenarios).


zptro:
@atte-wsp, you do not have to review this, but I thought you might be interested.

johpiip (approved):

PR #355, opened 2021-08-23
Enable workforce accessibility calculation and printing

zptro:
The "sop" (as well as the no longer existant "so") model has no `AccessibilityModel`, so the accessibility calculation needs to be disabled for it. By accident, I had disabled it for workforce accessibility `Purpose` as well. This PR fixes the problem. 

However, with the `AccessibilityModel` refactoring done, workforce accessibility calculation could maybe be made more streamlined with an own subclass.


johpiip (approved):
Tested with test data and mock assignment and it works!

PR #357, opened 2021-08-25
Print KPI summary

zptro:
- New result file "result_summary.txt" with aggregate results and key performance indicators.
- Whole model area aggregation to area average aggregation prints
- Workplace effective density printed to same file as workforce accessibility


johpiip (dismissed):

johpiip:
Oops, I did not refresh so I approved. 😆 Sorry!

Scripts/utils/zone_interval.py
```diff
@@ -140,6 +140,7 @@ def averages(self, array, weights):
             Aggregated array
         """
         aggregation = pandas.Series(index=self.keys)
+        aggregation["all"] = numpy.average(array, weights=weights)
```
zptro, line 143: When `array` and `weights` come from the logit model, they are usually only for "HS15" area. So the "all" aggregation will in this case be for "HS15". The "peripheral" aggregation will in these cases be zero.

johpiip: I see here a chance of misunderstandings... 😕

johpiip (approved):

PR #366, opened 2021-08-31
Fix bus kms

### LINKED ISSUE #365: Bus sums in result files are not the same
- transit_kms.txt bus sum is 415 966 km	
- vehicle_kms_vdfs.txt "bus" sum 415 869 km
  - Should we have and "other" class?
- vehicle_kms_areas.txt "bus" sum 
  - model area 503 376 km		
  - HS15 area 423 153 km
  - This is all transit modes, not only bus

zptro:
Closes #365.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -483,20 +483,21 @@ class AssignmentPeriod(Period):
         # calc @bus and data3
         heavy = (self.extra("truck"), self.extra("trailer_truck"))
         for link in network.links():
-            freq = 0
-            for segment in link.segments():
-                segment_hdw = segment.line[self.extra("hw")]
-                if 0 < segment_hdw < 900:
-                    freq += 60 / segment_hdw
-            link[self.extra("bus")] = freq
-            if link.type // 100 in param.bus_lane_link_codes[self.name]:
-                # Bus lane or rail
-                link[background_traffic] = 0
-            else:
-                link[background_traffic] = freq
-            if include_trucks:
-                for ass_class in heavy:
-                    link[background_traffic] += link[ass_class]
+            if link.type > 100: # If car or bus link
+                freq = 0
+                for segment in link.segments():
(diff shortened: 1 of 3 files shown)


Scripts/assignment/emme_assignment.py
```diff
@@ -154,6 +154,7 @@ def aggregate_results(self, resultdata):
         # Aggregate and print vehicle kms and link lengths
         vdfs = {param.roadclasses[linktype].volume_delay_func
             for linktype in param.roadclasses}
+        vdfs.add(0)
```
johpiip, line 157: This looks a bit funny to me. Is there some other way to add "links without car traffic (= without volume delay function)" than manually?

Scripts/assignment/assignment_period.py
```diff
@@ -483,20 +483,21 @@ def _calc_background_traffic(self, include_trucks=False):
...
-            link[self.extra("bus")] = freq
-            if link.type // 100 in param.bus_lane_link_codes[self.name]:
-                # Bus lane
-                link[background_traffic] = 0
-            else:
-                link[background_traffic] = freq
-            if include_trucks:
-                for ass_class in heavy:
-                    link[background_traffic] += link[ass_class]
+            if link.type > 100: # If car or bus link
```
johpiip, line 486: Will link types 0-100 now be considered as vdf=0?

johpiip (approved):

Scripts/assignment/assignment_period.py
zptro: No, not because of this code. This code forces attribute `@bus` to zero for rail links.

johpiip (approved):
Sorry, I had missed that this had to be re-reviewed! I approved this. Related to this, I was wondering last weekend about the naming convention for `0` class. If we want it to be other than `0`, I think `no_vdf` is better than `missing` or `other`. Especially `missing` implies that there should be a vdf for these links even if there really shouldn't.

PR #367, opened 2021-09-01
Remove 999999 replacements for time values

zptro:
The code has looped through all OD-impedances (including travel time) and replaced them by 999999 if travel time is greater than 999999 (i.e., no path was found). In python 2.7 this worked fine, because travel time was usually last in the dict, so that replacement happened last. In python 3.7 however, travel time is usually first in the dict (because it was inserted first), so when handling the other impedances, travel time would no longer be greater than 999999!

I see not reason to replace travel time by 999999, as it will be a very large number anyway.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -134,10 +134,12 @@ class AssignmentPeriod(Period):
 
         mtxs = {imp_type: self._get_emmebank_matrices(imp_type, iteration=="last")
             for imp_type in ("time", "cost", "dist")}
-        # fix the emme path analysis results (dist and cost zero if path not found)
-        for mtx_type in mtxs:
+        # fix the emme path analysis results
+        # (dist and cost zero if path not found)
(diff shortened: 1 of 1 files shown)


johpiip (approved):
I am approving but I do not really understand the PR message versus the changeset. 😕 As far as I see, the changeset contains a) looping explicitly through "cost" and "dist" matrices in that order, and b) defining "path not found" as *greater than or equal to* 999999 instead of *greater than* 999999. Do you mean that if "cost" and "dist" matrices change, also "travel time" matrix changes consequently?

zptro:
The change from `>` to `=>` was actually a mistake, good that you noticed! 😄 

Hence, (a) is the point of this PR. "time" matrices should not be looped through, as that messes up what we are trying to achieve, at least in python 3.7.

zptro:
Changing `>` to `>=` was an alternative way of solving this that I had in mind. But equality checks are unstable for floats, so I ditched that idea.

johpiip (approved):
Now I get it, thanks! :)

PR #371, opened 2021-09-15
Convergence indicators

zptro:
Print two demand model convergence indicators.

The next step could be to make it possible to set a convergence threshold instead of fixed iteration number, but this requires some investigations on what the default value could be.


johpiip (approved):
Nice!

PR #373, opened 2021-09-21
Insert node result aggregation into loop

zptro:
@MerviV found that node boarding results for whole day were always zero. As it turned out, this aggregation was by accident not inside the loop iterating over result types.


johpiip (approved):
Nice catch! Python and its whitespace is always tricky. 👍

PR #374, opened 2021-09-21
Unify connector handling

zptro:
Give connectors an own `RoadClass` so they do not have to be treated as exceptions.


Scripts/assignment/assignment_period.py
```diff
@@ -288,14 +286,12 @@ def _set_car_and_transit_vdfs(self):
             elif linktype in param.custom_roadtypes:
                 # Custom car link
                 link.volume_delay_func = linktype - 90
-                for linktype in range(l, u):
+                for linktype in param.roadclasses:
```
zptro, line 289: In python 3.7+, dicts are always iterated in insertion order, so the separate range (which does not work if link type numbers are directly following each other) is unnecessary.

johpiip (approved):

PR #375, opened 2021-09-29
Add central transit accessibility calculation

zptro:
In MALPAKKA analyses, central transit accessibility is used as an alternative accessibility indicator to standard logsum accessibility. It is the combination of travel time and cost to zone 101 (i.e., the first zone).


hsl-petrhaj:
Inactive pull request.

PR #376, opened 2021-10-01
Hotfix/headways

zptro:
Fix congested transit assignment by writing time-period-specific headways to `headway` attribute, because this attribute is used for calculating line capacities.

### DIFF
--- a/Scripts/assignment/emme_bindings/mock_project.py
+++ b/Scripts/assignment/emme_bindings/mock_project.py
@@ -309,7 +309,7 @@ class MockProject:
             "TRANSIT_SEGMENT", "@base_timtr", "", 1.0,
             overwrite=True, scenario=scenario)
         report = {
-            "stopping_criterion": "MAX_ITERATIONS",
+            "stopping_criteria": "MAX_ITERATIONS",
             "iterations": [{"number": 1}],
         }
         return report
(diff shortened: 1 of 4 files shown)


Scripts/assignment/emme_bindings/mock_project.py
```diff
@@ -309,7 +309,7 @@ def congested_assignment(self, transit_assignment_spec, class_names,
             "TRANSIT_SEGMENT", "@base_timtr", "", 1.0,
             overwrite=True, scenario=scenario)
         report = {
-            "stopping_criterion": "MAX_ITERATIONS",
+            "stopping_criteria": "MAX_ITERATIONS",
```
johpiip, line 312: "criterion" is singular and "criterion" is plural, so this change should not be needed? I did a quick Find All for `olusanya` and both are used...

johpiip (approved):

Scripts/assignment/emme_bindings/mock_project.py
zptro: Yes, but the most important task of `MockProject` is to produce the exact same result as EMME. 😄

johpiip: Ah okay, if it is in Emme then nothing! 😄

PR #394, opened 2021-11-10
Print zone-specific CBA (consumer) results to text file

zptro:
- travel time change
- travel cost change
- distance change
- revenue change

### DIFF
--- a/Scripts/cba.py
+++ b/Scripts/cba.py
@@ -171,10 +171,10 @@ def run_cost_benefit_analysis(scenario_0, scenario_1, year, workbook):
     Parameters
     ----------
     scenario_0 : str
-        Name of do-nothing scenario, for which 
+        Name of do-nothing scenario, for which
         forecast results are available in Results folder
     scenario_1 : str
-        Name of project scenario, for which 
+        Name of project scenario, for which
         forecast results are available in Results folder
     year : int
         The evaluation year (1 or 2)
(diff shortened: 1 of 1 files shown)


Scripts/cba.py
```diff
@@ -395,13 +412,17 @@ def calc_revenue(demands, costs):
...
     log.initialize(args)
     wb = load_workbook(os.path.join(SCRIPT_DIR, "CBA_kehikko.xlsx"))
-    run_cost_benefit_analysis(
+    results = run_cost_benefit_analysis(
         args.baseline_scenario, args.projected_scenario, 1, wb)
-    if args.baseline_scenario_2 is not None and args.baseline_scenario_2 != "undefined":
+    if (args.baseline_scenario_2 is not None
+            and args.baseline_scenario_2 != "undefined"):
         run_cost_benefit_analysis(
             args.baseline_scenario_2, args.projected_scenario_2, 2, wb)
```
johpiip, line 420: What about the results in this second CBA analysis?

johpiip (approved):

Scripts/cba.py
```diff
@@ -240,33 +248,42 @@ def run_cost_benefit_analysis(scenario_0, scenario_1, year, workbook):
         revenues_car = 0
         cols = CELL_INDICES["gains"]["cols"]
         rows = CELL_INDICES["gains"]["rows"][year]
-        for tc in param.transport_classes:
+        for transport_class in param.transport_classes:
```
johpiip, line 251: Good clarification!

Scripts/cba.py
```diff
@@ -231,7 +238,8 @@ def run_cost_benefit_analysis(scenario_0, scenario_1, year, workbook):
         ws[indices[mode]] = station_diff["number"][mode]
 
     # Calculate gains and revenues
-    for tp in ["aht", "pt", "iht"]:
+    results = defaultdict(float)
+    for timeperiod in ["aht", "pt", "iht"]:
```
johpiip, line 242: Good clarification!

johpiip (approved):

Scripts/cba.py
zptro: I think we can neglect them. Would need to rethink the whole structure, because per-day results for two different years cannot just be summed directly, they would need to be discounted.

PR #405, opened 2021-11-19
Update README.md

zptro:
Updated the documentation on `dev-config.json`. Also some small improvements in general README.

### DIFF
--- a/Scripts/README.md
+++ b/Scripts/README.md
@@ -111,7 +111,7 @@ These should not be used when running model system from command line!
 Instead `helmet.py` flag parameters should be used.
 These can be set if model system is run from UI, to set parameters that cannot be set in UI.
 A flag is activated by putting its name inside the brackets,
-flags are separated by commas. 
+flags are separated by commas (e.g., `"OPTIONAL_FLAGS": ["RUN_AGENT_SIMULATION", "DO_NOT_USE_EMME"]`).
 
 #### `END_ASSIGNMENT_ONLY`
 
--- a/Scripts/README.md
+++ b/Scripts/README.md
@@ -8,6 +8,11 @@ which can be used to setup the model run in advance.
 
 ## Configuring the model run with `dev-config.json`
 
+### `HELMET_VERSION`
+
+The version number is logged in model runs.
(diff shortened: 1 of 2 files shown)


Scripts/README.md
```diff
@@ -108,9 +94,52 @@ some zone a zero value line instead of a missing line must be used.
...
+
+### `REL_GAP`
+
+Convergence criterion: Car work matrix relative change between iterations.
+
+### `OPTIONAL_FLAGS`
+
+These should not be used when running model system from command line!
+Instead `helmet.py` flag parameters should be used.
+These can be set if model system is run from UI, to set parameters that cannot be set in UI.
```
johpiip, line 112: What do you mean by these instructions?

There are three ways to run the model with parameters, right?
1. UI
2. CLI with parameters in the command line call e.g. `python helmet.py --scenario-name awesomescenario`
3. CLI with parameters in the `dev-config.json` when the command line call is just `python helmet.py`

Should this say that optional flags should not be used in `dev-config.json` so that they do not override UI parameters?

Scripts/README.md
```diff
@@ -108,9 +94,52 @@ some zone a zero value line instead of a missing line must be used.
...
+A flag is activated by putting its name inside the brackets,
+flags are separated by commas.
```
johpiip, line 114: I'd add here an example:

```
"OPTIONAL_FLAGS": ["DO_NOT_USE_EMME"]
```

Scripts/README.md
johpiip: Oops, I did not realize we were at `dev-config.json` manual. 😄 I'm going to leave this here anyway but it is not that relevant anymore.

johpiip (approved):
Hey, just wondering, what is the use case for these model system manuals when we have the [helmet-ui manual](https://hsldevcom.github.io/helmet-ui/) too? Should they be kept separate or should we try to merge them?

Scripts/README.md
zptro: Good suggestion!

johpiip (approved):
Nice!

PR #406, opened 2021-11-22
Add standard attributes to link attribute dump

zptro:

PR #409, opened 2021-11-23
Set path properly instead of re-initializing `ResultData` object

zptro:
`ResultData` object stores the result data path. However, the previous fix #402 destroyed this link by creating a new object. Now the path is properly updated, while also creating necessary folders.


johpiip:
I have not yet tested this. I ran the previous `mal23` runs on `feat/print_standard_attr` #406.

johpiip (approved):
Works when I ran this with test data!

PR #412, opened 2021-11-29
Fix bike mode and refactor mode handling

zptro:
By mistake, bike has been assigned to car mode: https://github.com/HSLdevcom/helmet-model-system/pull/290/commits/bd716bd18eadf232ed9d712f8d3e1ee0b3d045f3#diff-592ba4818568720b71888a1390079ea0f3ed8650cbd11e11d33330e0757f9d3bR365
The idea was to assign it to the main `AUTO` mode (which temporally in the previous PR was `c`, but now is `h`).

This PR puts bike assignment back to mode `f`. The main `AUTO` mode is still `h` in bike assignment, but that has no practical implication other than that `h` must also be allowed on the bike links in bike assignment.

At the same time I moved out hard-coded `h`, `f` and `c` from the code, to avoid similar bugs in the future.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -277,6 +277,8 @@ class AssignmentPeriod(Period):
         network = self.emme_scenario.get_network()
         transit_modesets = {modes[0]: {network.mode(m) for m in modes[1]}
             for modes in param.transit_delay_funcs}
+        main_mode = network.mode(param.main_mode)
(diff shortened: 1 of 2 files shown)


annika-wsp (commented):
I'm not so familiar with the code that I could find anything wrong with the fix. I'll test it when a new release is out.

johpiip (approved):
Interesting!

PR #423, opened 2022-01-04
Fix git tag

zptro:
This fixes two problems:
1. When run from UI, the os directory is the path of the UI, not the model system. So we need to manually change it to the model system directory to be able to detect git tags.
2. `subprocess.check_output` automatically sends error messages to `stderr`, even if the exception is caught. We can avoid this by sending the error message to `subprocess.STDOUT`, which in this case is the pipe leading to the return value of this function.


johpiip (approved):
Nice!

PR #426, opened 2022-01-05
Fix node number validation

zptro:
The previous method could not handle modes that are allowed in several node number intervals. The new method checks all modes, not only transit modes.

### DIFF
--- a/Scripts/parameters/assignment.py
+++ b/Scripts/parameters/assignment.py
@@ -82,13 +82,14 @@ transit_delay_funcs = {
         "iht": 6,
     },
 }
-# Node numbers used in HSL official networks
+# Node numbers used in HSL official networks and their allowed modes
 official_node_numbers = {
-    "bgde": (40000, 599999),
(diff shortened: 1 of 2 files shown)


johpiip (approved):
Interesting approach! Are there tests written on these validation procedures?

zptro:
> Interesting approach! Are there tests written on these validation procedures?

Yes, the network validation is tested here: https://github.com/HSLdevcom/helmet-model-system/blob/39b57ecbbd541ab993a6e5ff07dd08f66d6c5390/Scripts/tests/unit/test_assignment.py#L31-L33

Maybe it should be separated into into an own test for clarity.

johpiip:
Ah, okay, thanks! I just thought that if a new validation procedure was added here, there could also be a new test case to test it.

zptro:
> Ah, okay, thanks! I just thought that if a new validation procedure was added here, there could also be a new test case to test it.

Yes, you are right. Overall, tests should be improved and actually test different cases.

PR #427, opened 2022-01-05
Add land-use scenarios

zptro:


johpiip (approved):
Looks fine by me! I did not run anything though. Maybe `haj` and `kes` test files could be different from one another as real data would be but that is not very important.

PR #437, opened 2022-02-04
Add text file dumping of link attributes with wkt coordinates

zptro:
Extracted from #360 and #406, which were part of dropped branch mal23.

If we do not want this as a part of model runs, we could implement it as a Modeller toolbox, similar to the network validation toolbox.


johpiip (approved):
Great! I'm approving this but I have not done a model run to test it! :(

PR #438, opened 2022-02-04
Bring car ownership arguments to remote work scenario

johpiip:
We noticed that some sensitivity anmalysis scenarios need both remote work generation and car ownership changes enabled.


zptro (approved):
The `mal23-muulitar` branch fails integration test (because it alters results!), but that does not have anything to do with this PR.

johpiip:
@zptro Ah yes, true!

PR #440, opened 2022-02-10
Link print toolbox

zptro:
Modeller toolbox for simple link attribute file dump.

Includes noise calculation (works well only when morning peak hour is assigned in the same EMME scenario). Noise calculation could be removed from here if noise extra attribute would be added.


Scripts/utils/calc_noise.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Parameters
+    ----------
+    morning_network : inro.emme.network.Network
+        Network used for morning peak hour assignment
+    light_modes : tuple of str
+        Extra attribute names for 24-h traffic per assignment class
+    heavy_modes : tuple of str
+        Extra attribute names for 24-h traffic per assignment class
+    """
```
johpiip, line 16: If you have a source for these noise calculations formulas, this would be a great place to cite it. :)

johpiip (approved):
Great! Look good to me, I did not test it in production though!

PR #443, opened 2022-02-16
Re-arrange summary sheet in cba excel file

zptro:
Changed layout of cba excel summary sheet to better resemble what is usually presented in transport investment appraisals.

~~The data bars are extra candy.~~ The data bars shown in the picture had to be removed, because `openpyxl` mess them up.

Kuva1


johpiip (approved):
Looks fancy! I guess all the changes were either cosmetical or on cells that only reference other cells? That's why there are no Python changes?

zptro:
> Looks fancy! I guess all the changes were either cosmetical or on cells that only reference other cells? That's why there are no Python changes?

Yes, all cells on summary sheet reference other cells.

PR #450, opened 2022-03-14
Fix/noise zones

### LINKED ISSUE #449: Modeller exception: LinkPrinting run ended with exception
When using Print links from HELMET toolbox on ve0:

Modeller Exception
LinkPrinting run ended with exception: local variable 'zone_width' referenced before assignment

johpiip:
Fixes #449

### DIFF
--- a/Scripts/utils/calc_noise.py
+++ b/Scripts/utils/calc_noise.py
@@ -75,4 +75,6 @@ class NoiseModel:
             if interval[0] <= start_noise < interval[1]:
                 zone_width = func[interval](start_noise - interval[0])
                 break
+        else:
+            raise ValueError("{}".format(link.id))
         return zone_width
--- a/Scripts/utils/calc_noise.py
+++ b/Scripts/utils/calc_noise.py
@@ -75,5 +75,5 @@ class NoiseModel:
                 zone_width = func[interval](start_noise - interval[0])
                 break
         else:
-            raise ValueError("{}".format(link.id))
+            raise ValueError("No noise interval found for link {}".format(link.id))
         return zone_width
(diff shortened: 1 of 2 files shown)


Scripts/utils/calc_noise.py
```diff
@@ -75,4 +75,6 @@ def calc_noise(self, link):
             if interval[0] <= start_noise < interval[1]:
                 zone_width = func[interval](start_noise - interval[0])
                 break
+        else:
+            raise ValueError("{}".format(link.id))
```
zptro, line 79:
```suggestion
            raise ValueError("No noise interval found for link {}".format(link.id))
```

zptro (changes requested):
The problem could be solved by setting `traffic` to at least 0.01 here:

https://github.com/HSLdevcom/helmet-model-system/blob/f43f167ca5caae9aa2ee776e9b35c9b46d19a37f/Scripts/utils/calc_noise.py#L36

instead of here:

https://github.com/HSLdevcom/helmet-model-system/blob/f43f167ca5caae9aa2ee776e9b35c9b46d19a37f/Scripts/utils/calc_noise.py#L46

Then ´cross_traffic` would also be at least 0.01, which would prevent noise from going below zero.

zptro (approved):

johpiip:
Nice! I forgot about this when I had to hurry assessments but great that you did not. :) Should we make a test run or did you do that already?

zptro:
> Nice! I forgot about this when I had to hurry assessments but great that you did not. :) Should we make a test run or did you do that already?

Yes, we should probably do a 2040 test run.

johpiip:
@zptro I tested this and it works! Those links which had a noise zone width of 5 still have it. Ve0 scenario does not crash when writing links.

PR #453, opened 2022-05-16
Check that workplace shares are not larger than one

zptro:
This halts the model run if workplace shares larger than one are detected. As MAL 2019 data contains such anomality, we should either wait with pulling this or somehow change it into a warning.


johpiip (approved):
We could change `log.error` to `log.warn` [here](https://github.com/HSLdevcom/helmet-model-system/blob/8f7b36cead5724130677997f5912547a398f2605/Scripts/datahandling/zonedata.py#L238) but it might deteriorate land-use data quality if people do not notice warnings.

PR #459, opened 2022-09-02
Remove hard-coded area-based bounds from `Purpose`

zptro:
Introduce more general sub-bounds that can define separate parameters for arbitrary number of sub-regions instead of two.


Scripts/parameters/zone.py
```diff
@@ -93,6 +93,12 @@
         "area": "all",
     },
 )
+purpose_areas = {
+    "metropolitan": (0, 6000, 16000),
+    "peripheral": (16000, 31000),
+    "all": (0, 6000, 31000),
+    "external": (31031, 40000),
+}
```
johpiip, line 101: Consider adding a helpful comment here because it is difficult to understand why some areas have three items and some only two.

johpiip (dismissed):
Thank you for the PR! I understand what you are going for but the code is too complex for me to review. Tests are passing so I guess that is enough. I added one comment on the `purpose_areas` variable!

johpiip (approved):

zptro:
Everything seems good here, but I am not authorized to merge, @johpiip.

johpiip:
> Everything seems good here, but I am not authorized to merge, @johpiip.

Sorryyy, I did not realize! 😢

PR #463, opened 2022-09-15
Fix bicycle trips CBA

### LINKED ISSUE #461: Fix bicycle trips in CBA
Cell reference points to wrong variable.

samakinen:
Fix incorrect linking of bicycle trips in CBA Excel sheet that causes the travel time saving of leisure trips being calculated based on the work trip data.
Fixes #461


johpiip (approved):
Thanks, I checked that Excel looks good!

PR #464, opened 2022-09-19
Fix travel times for heavy modes when there are toll costs on the network

### LINKED ISSUE #462: Toll cost is written to result matrices for trucks
Truck and trailer truck assignment is not affected by toll cost. However, in this line we still calculate toll cost as if they would be affected. This later on affects the travel time matrix calculation, as that is calculated by subtracting the cost matrix from the generalized cost matrix.

https://github.com/HSLdevcom/helmet-model-system/blob/4939e0505b8bde26f53116c89c31ff747cf74c45/Scripts/assignment/datatypes/car.py#L27

johpiip:
Fixes #462


Scripts/assignment/assignment_period.py
```diff
@@ -476,7 +476,10 @@ def _extract_timecost_from_gcost(self, ass_class):
         gcost = self._get_matrix("gen_cost", ass_class)
         cost = self._get_matrix("cost", ass_class)
         dist = self._get_matrix("dist", ass_class)
-        time = gcost - vot_inv*(cost + self.dist_unit_cost*dist)
+        time = gcost - vot_inv*self.dist_unit_cost*dist
```
zptro, line 479: In truck assignment, `vot_inv*self.dist_unit_cost` is not used as-is due to historical reasons; in Helmet 2 it used to be hard-coded as 0.2 min/km and no one has thought of changing it since, so it is still hard-coded here: 

https://github.com/HSLdevcom/helmet-model-system/blob/4939e0505b8bde26f53116c89c31ff747cf74c45/Scripts/assignment/datatypes/car_specification.py#L40-L42

What we also *should* do is to define an own `dist_unit_cost` for trucks and trailer trucks and replace 0.2 above by `vot_inv*dist_unit_cost`. This distance unit cost should presumably be a bit higher than for private cars. This solution will change route choices somewhat.

A quick fix to get `time` matrix calculation right for trucks and trailer trucks would be to set `dist_unit_cost` for trucks and trailer trucks so that `vot_inv*self.dist_unit_cost` results in 0.2.

johpiip:
Cost matrices probably shouldn't change between a scenario with toll costs and a scenario with no toll costs... Changing this to a draft while fixing.

zptro (changes requested):
This is an elegant solution! Unfortunately it will not give very accurate results, due to my weak coding elsewhere.

zptro:
> Cost matrices probably shouldn't change between a scenario with toll costs and a scenario with no toll costs... Changing this to a draft while fixing.

I think this is ok! When private cars leave tolled routes, trucks can sometimes start preferring these, as they get faster.

johpiip:
> > Cost matrices probably shouldn't change between a scenario with toll costs and a scenario with no toll costs... Changing this to a draft while fixing.
> 
> I think this is ok! When private cars leave tolled routes, trucks can sometimes start preferring these, as they get faster.

Hmm, I might have understood something wrong then. What I thought was that cost_aht.omx (and pt and iht) for trailer trucks and trucks should always be 0 or 999999. At least, that's what it is in scenarios with no toll costs. And if toll costs do not affect trailer trucks and truck, then cost matrices should be 0 or 999999 also in those scenarios.

zptro:
> > > Cost matrices probably shouldn't change between a scenario with toll costs and a scenario with no toll costs... Changing this to a draft while fixing.
> > 
> > 
> > I think this is ok! When private cars leave tolled routes, trucks can sometimes start preferring these, as they get faster.
> 
> Hmm, I might have understood something wrong then. What I thought was that cost_aht.omx (and pt and iht) for trailer trucks and trucks should always be 0 or 999999. At least, that's what it is in scenarios with no toll costs. And if toll costs do not affect trailer trucks and truck, then cost matrices should be 0 or 999999 also in those scenarios.

Hmm, you are right. In the final matrices they should always be 0 or 999999. I messed this up, because in the iterations before the last iteration, where cost matrices are fed into demand models, distance cost is part of cost matrices.

zptro:
So it seems you discovered another bug. This row is applied for truck and trailer trucks, even if the that extra attribute does not affect route choices!

https://github.com/HSLdevcom/helmet-model-system/blob/4939e0505b8bde26f53116c89c31ff747cf74c45/Scripts/assignment/datatypes/car.py#L27

zptro:
> So it seems you discovered another bug, @johpiip. This row is applied for truck and trailer trucks, even if the that extra attribute does not affect route choices!
> 
> https://github.com/HSLdevcom/helmet-model-system/blob/4939e0505b8bde26f53116c89c31ff747cf74c45/Scripts/assignment/datatypes/car.py#L27

Or actually this was the original bug that I had identified in the issue. :grin:

Scripts/assignment/assignment_period.py
johpiip: Thank you for the details! Now I think I got it!

Scripts/parameters/assignment.py
```diff
@@ -183,6 +183,8 @@
 }
 # Default distance unit cost [eur/km]
 dist_unit_cost = 0.12
+# Default speed for trucks and trailer trucks [min/km]
```
zptro, line 186: Maybe a bit confusing to talk about "speed", when this describes the trade-off between time and distance.

Scripts/assignment/assignment_period.py
```diff
@@ -476,7 +476,11 @@ def _extract_timecost_from_gcost(self, ass_class):
         gcost = self._get_matrix("gen_cost", ass_class)
         cost = self._get_matrix("cost", ass_class)
         dist = self._get_matrix("dist", ass_class)
-        time = gcost - vot_inv*(cost + self.dist_unit_cost*dist)
+        if ass_class not in ("trailer_truck", "truck"):
```
zptro, line 479: By flipping the `if` and `else`, we could get rid of the `not`, but that is not so important.

zptro (approved):
Nice work! Should the cost matrices for truck and trailer truck be set to zero?

Scripts/assignment/assignment_period.py
```diff
@@ -142,6 +142,10 @@ def assign(self, matrices, iteration):
                 path_not_found = mtxs["time"][mtx_class] > 999999
                 mtxs[mtx_type][mtx_class][path_not_found] = 999999
         # adjust impedance
+        if iteration == "last":
+            for mtx_class in ("trailer_truck", "truck"):
+                path_not_found = mtxs["time"][mtx_class] <= 999999
+                mtxs["cost"][mtx_class][path_not_found] = 0
```
zptro, line 148: Why not simply like this? I think this will give the same result.

```suggestion
                mtxs["cost"][mtx_class] = 0
```

johpiip: Because before, there was 999999s in the cost matrices despite the comment in L139: https://github.com/HSLdevcom/helmet-model-system/blob/6b3d1ebd1ead1f464ef8fecdc5f4a13c460433c7/Scripts/assignment/assignment_period.py#L139-L143 So I mimiced the behavior.

zptro: I see.

The comment in L139 is actually a bit confusing, as it describes the problem and not the solution. The solution is that dist and cost is set to 999999 if no path is found.

In your code it is actually the bool matrix name `path_not_found` that is confusing. According to your definition, it represents the opposite (i.e., `path_found`).

zptro: I *think* a more elegant solution could be to initialize the impedance matrices to 0 here:

https://github.com/HSLdevcom/helmet-model-system/blob/6b3d1ebd1ead1f464ef8fecdc5f4a13c460433c7/Scripts/assignment/emme_assignment.py#L86-L91

But I am not 100 % sure that it does not have any side effects.

johpiip: Ooops, you are totally right on the naming, I though I fixed it but I did not. 🙈 I made a new commit to fix that and docs.

Changing the default looks risky!

zptro (approved):
The truck and trailer truck exception in toll cost handling has never been elegant, but now it is at least consistent and well-documented. Excellent job, @johpiip!

PR #469, opened 2022-09-26
Generalize sub-bound handling

zptro:
#459 generalized sub-region parameter handling. Now I found some places that I had overlooked. With this PR sub-region parameters will be fully flexible in agent model as well.

Also fixes zone interval for external demand.


johpiip (approved):

johpiip (approved):
I tested by running 2018 and it works! Results are identical to those before bugs. :) Thank you! I am merging this.

PR #479, opened 2022-12-22
Add mode validation

zptro:
Add feature to netowork validation: Check that the modes defined in assignment parameters are actually found in the network.

To be able to test this feature, mock assignment module needed some extensions.


johpiip:
Can you explain the purpose of this PR, thanks! :)

zptro:
> Can you explain the purpose of this PR, thanks! :)

Sorry, now I have added a short description.

Scripts/utils/validate_network.py
```diff
@@ -42,6 +42,13 @@ def validate(network, fares=None):
                 found_zone_share)
             log.error(msg)
             raise ValueError(msg)
+    validate_mode(network, param.main_mode, "AUTO")
```
johpiip, line 45: Are "AUTO", "AUX_AUTO", "TRANSIT", and "AUX_TRANSIT" here names that come from how Emme API reads networks? Consider moving these magic variables as a parameter.

johpiip (commented):
I can't do any tests right now because I have model runs going on *both* of our Emme machines, so I will continue next week. 😄

Scripts/utils/validate_network.py
zptro: Yes, they come from the Emme text file import syntax. Now I moved the strings to the beginning of the file as constants.

johpiip (approved):
Awesome, thanks!

PR #483, opened 2023-01-18
Change model run defaults

johpiip:
MAL 2040 ve0 and ve2 model runs need higher counts on both car assignment and demand model iterations to fully converge. This will add run time only on those model runs which have trouble converging. Other model runs (eg. 2018 or 2023) converge early and do not take extra time.


eevavesaoja (approved):
Hyvältä näyttää!

PR #485, opened 2023-02-08
Fix headways in extra waiting times

### LINKED ISSUE #484: Resuls are different in first and second model run
Results are different when a network is run the first and the second time. The 2nd, 3rd, and 4th run are identical with each other.

Using:
- helmet-model-system v4.1.2
- Emme 4.6.2
- test network and land-use (but also 2018, not so many times tested and verified though)

johpiip:
Fixes a bug where headways in function `_calc_extra_wait_time()` were read from Emme networks instead of extra attributes. This caused the first model run to have different results than the following model runs, because the first run reads them from `transit_lines_xxx.txt` and other runs read them from the last congested transit assignment i.e. `@hw_iht`: https://github.com/HSLdevcom/helmet-model-system/blob/089f9c43113a97aaf5e9e975887eddf07ebe2330/Scripts/assignment/assignment_period.py#L766-L769

Fixes #484

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -743,7 +743,8 @@ class AssignmentPeriod(Period):
                                   + b["ctime"]*cumulative_time
                                   + b["cspeed"]*cumulative_speed)
                 # Estimated waiting time addition caused by headway deviation
-                segment["@wait_time_dev"] = headway_sd**2 / (2.0*line.headway)
+                headway_attr = self.extra("hw")
+                segment["@wait_time_dev"] = headway_sd**2 / (2.0*line[headway_attr])
         self.emme_scenario.publish_network(network)
 
     def _assign_transit(self):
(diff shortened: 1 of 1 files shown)


Scripts/assignment/assignment_period.py
```diff
@@ -743,7 +743,8 @@ def _calc_extra_wait_time(self):
                                   + b["ctime"]*cumulative_time
                                   + b["cspeed"]*cumulative_speed)
                 # Estimated waiting time addition caused by headway deviation
-                segment["@wait_time_dev"] = headway_sd**2 / (2.0*line.headway)
+                headway_attr = self.extra("hw")
```
zptro, line 746: Consider lifting definition of `headway_attr` out of `segment` loop.

zptro (dismissed):
Very nice catch! And embarrassing for me that this bug has been there for a few years.

eevavesaoja (dismissed):
Thank you for fixing this!

Scripts/assignment/assignment_period.py
johpiip: Thanks, now fixed!

johpiip:
Thanks and don't worry @zptro! Should I make a similar PR to `master` for v4.0 or is it considered to be at its end-of-life?

zptro (approved):

zptro:
> Thanks and don't worry @zptro! Should I make a similar PR to `master` for v4.0 or is it considered to be at its end-of-life?

Good question. I am probably not the one to decide this, but maybe it can be considered EOL.

PR #489, opened 2023-02-14
Rewrite `DemandModel` code

zptro:
Module `demand/trips` has been relatively untouched since I originally wrote it four years ago. Now I thought it would be a good time to go through the code and improve it.
- Move age group definition to parameter file
- Define model bounds only in one place
- Clarify code


johpiip (approved):
Looks good and the tests are passing! :)

PR #490, opened 2023-02-28
Add scenario id when setting and getting emme matrices

zptro:
@samakinen suggested this change, which prohibits sudden crashes when emme scenarios have different number of zones.

I do not think it is good practice to mix different zone systems (and hence differently sized matrices) in the same emme bank, but I admit that it might sometimes be handy to be able to try small zone splittings in an emme scenario without having to create an entirely new emme bank. And most importantly, the system should not crash unexpectedly because of such a choice.

### DIFF
--- a/Scripts/assignment/assignment_period.py
+++ b/Scripts/assignment/assignment_period.py
@@ -411,10 +411,12 @@ class AssignmentPeriod(Period):
             raise ValueError(msg)
         elif result_type is None:
             self.emme_project.modeller.emmebank.matrix(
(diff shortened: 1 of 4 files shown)


johpiip (dismissed):
I have been there before - trying to run a test network in a real 2018 Emme bank. 😄 Your advice sounds good so maybe consider adding a warning like: "Running a model with xxx zones when earlier full matrices in the bank imply yyy zones."

johpiip (approved):
Awesome, thanks!

PR #494, opened 2023-03-14
Keep correct order in link_lengths.txt

hsl-petrhaj:
Changed linktypes to list, because list preserves the order of items. Flipped road and rail links.

### DIFF
--- a/Scripts/assignment/emme_assignment.py
+++ b/Scripts/assignment/emme_assignment.py
@@ -165,11 +165,13 @@ class EmmeAssignmentModel(AssignmentModel):
         area_kms = {ass_class: pandas.Series(0.0, areas)
             for ass_class in ass_classes}
         vdf_area_kms = {vdf: pandas.Series(0.0, areas) for vdf in vdfs}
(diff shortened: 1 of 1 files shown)


zptro (commented):
A one-liner solution that I found (https://stackoverflow.com/questions/7961363/removing-duplicates-in-lists): `linktypes = list(dict.fromkeys(param.roadtypes.values())) + list(dict.fromkeys(param.railtypes.values()))`

But it is a bit harder to read of course.

hsl-petrhaj:
@zptro Your version looks better, how can I resubmit the pull request with your changes included?

zptro:
> @zptro Your version looks better, how can I resubmit the pull request with your changes included?

You just commit the changes in your local branch and push to remote. Then this PR will also be updated.

johpiip (changes requested):
Good work! No need to add
> , Petr '23

because git blame will reveal that! :)

johpiip (approved):
Looks good to me! :)

PR #503, opened 2023-04-17
Update transit fare dtype

zptro:
In #495 I had forgot to change one `dtype` setting, because we have removed it in our Traficom fork of this repo.


hsl-petrhaj:
Looks reasonable.

hsl-petrhaj (approved):

PR #504, opened 2023-04-18
Reducing the network checking macro

hsl-petrhaj:
Reducing the macro by the functions already implementing in Python.


johpiip (commented):
Check with @MerviV that this works. 🙂

johpiip:
I will approve this once you have checked this with Mervi. 😊

PR #506, opened 2023-04-18
Pikaratikka is added in the mode checks

hsl-petrhaj:
Pikaratikka (light rail) was missing in the network checks


hsl-petrhaj:
@zptro or @johpiip please review

johpiip:
@hsl-petrhaj, you need to mark us as reviewers in the right tab. That gives us an email notice and we can start reviewing. 😊
kuva

hsl-petrhaj:
@johpiip I was wondering why is this not getting approved :D

zptro (approved):
Looks good to me!

PR #508, opened 2023-04-28
Expand time-period parameterization to mock_assignment and departure_time

zptro:
Expands the scope of #496


hsl-petrhaj (approved):
Looks good to me.

PR #509, opened 2023-05-02
Update second DepartureTimeModel call

zptro:
Fixes #508.


hsl-petrhaj (approved):
We could also consider writing some more tests in the future @zptro

zptro:
> We could also consider writing some more tests in the future @zptro

Indeed.

PR #511, opened 2023-05-09
Write volume-delay funcs using string formatting

zptro:


Scripts/parameters/assignment.py
```diff
@@ -91,38 +91,22 @@
     "htpaf": (802000, 806000),
     "hpaf": (810000, 816000),
 }
+vdf_temp = ("(put(60/ul2)*(1+{}*put((volau+volad)/{})/"
+            + "(ul1-get(2))))*(get(2).le.put(ul1*{}))*length+(get(2).gt."
+            + "get(3))*({}*get(1)*length+{}*(get(2)-get(3))*length)")
+buslane = "((lanes-1).max.0.8))"
```
hsl-petrhaj, line 97: @zptro I tested this part and there seems to be extra round bracket here.

zptro: yes, you are right. Sorry for that!

hsl-petrhaj (approved):
now good

PR #521, opened 2023-06-08
Emme network scripts

hsl-petrhaj:
Here I tried to address all the recent issues touching the network validation. I typically started by writing the test and then implemented the solution in network_validation.py. I am also publishing the tests because I think they might be useful in the future.
Some small reogranization in the folder structure is also proposed.


Scripts/utils/validate_network.py
```diff
@@ -61,8 +61,26 @@ def validate(network, fares=None):
         intervals += param.official_node_numbers[modes]
     unofficial_nodes = set()
     for link in network.links():
+        if len(link.modes) == 0:
```
zptro, line 64: This could be written as
```suggestion
        if not link.modes:
```

zptro (dismissed):
Very nice commit. Thank you for this!

Scripts/utils/validate_network.py
hsl-petrhaj: This should be in some "extreme Python" lessons :D

johpiip (dismissed):
Great error messages! :)

hsl-petrhaj:
Note: even our test network had some 999 links, I corrected these for some suitable close enough link type.

Scripts/utils/validate_network.py
```diff
@@ -99,7 +107,14 @@ def validate(network, fares=None):
...
                     log.error(msg)
                     raise ValueError(msg)
+                
+        if link.i_node.is_centroid and link.j_node.is_centroid:
+            msg = "Link {} is leading directly from centroid node {} to centroid node {}. This is not allowed.".format(link.id,link.i_node.number,link.j_node.number)
+            log.error(msg)
+            raise ValueError(msg)
+        
         for node in (link.i_node, link.j_node):
+
```
zptro, line 117: Why this extra row here?

zptro (approved):

PR #526, opened 2023-06-28
Add the measurement of convergence to the logging message

hsl-petrhaj:


hsl-petrhaj:
@zptro @johpiip
This introduces a message to inform the user about the convergence status. Can be picked up by the UI later.

zptro (commented):
Good thing to log, but could we not achieve the same thing by adding one single row (`log.info("Convergence fullfilled")`) here? https://github.com/HSLdevcom/helmet-model-system/blob/6305be17e0469530de6a58b25fbf1bd79e1a1bd5/Scripts/helmet.py#L116-L117

hsl-petrhaj:
> 

Good idea, it would have to with the else branch then? This elif only says when the criteria is not fulfilled.

zptro:
Maybe there should be a comment explaining the code. The elif states that convergence is achieved and that last assignment (with all its implications) should be run.

zptro:
The if-clauses could be rearranged a little bit if we also want to say that the demand has *not* converged.

zptro (approved):
Nice! If we want to fine-tune, we could make the condition in the two if-clauses into a variable and reuse it, but that is not a must.

hsl-petrhaj:
I have labeled the converge criteria under a variable with a fairly verbose name, but it should be easy to reuse and read this way.

zptro (approved):

PR #528, opened 2023-06-28
Removed scenario.id reference from if-clause where scenario is always None

hsl-petrhaj:


zptro (approved):
Nice fix!

I think we should do the same here:

https://github.com/HSLdevcom/helmet-model-system/blob/eec2f40f05de1005c929c0631f87e89dad536eb9/Scripts/helmet_validate_inputfiles.py#L163-L164

hsl-petrhaj:
Same fix added to the second occurence of the problem.

PR #531, opened 2023-06-29
Copy matrix folder from test scenario

hsl-petrhaj:
It's one small step for mock run, one giant leap for testing? Or not so much, but at least should enable having multiple testing scenarios under different names without having to move things in the folders manually.

### DIFF
--- a/Scripts/helmet_validate_inputfiles.py
+++ b/Scripts/helmet_validate_inputfiles.py
@@ -59,10 +59,10 @@ def main(args):
         mock_result_path = os.path.join(
             args.results_path, args.scenario_name, "Matrices")
         if not os.path.exists(mock_result_path):
-            msg = "Mock Results directory {} does not exist.".format(
-                mock_result_path)
-            log.error(msg)
-            raise NameError(msg)
+            log.info("Matrices folder does not exist, copying the folder from test scenario")
+            test_path = os.path.join(args.results_path,"test","Matrices")
+            import shutil
+            shutil.copytree(test_path, mock_result_path)
         assignment_model = MockAssignmentModel(MatrixData(mock_result_path))
         zone_numbers = assignment_model.zone_numbers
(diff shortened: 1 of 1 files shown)


Scripts/helmet_validate_inputfiles.py
```diff
@@ -59,10 +59,10 @@ def main(args):
         mock_result_path = os.path.join(
             args.results_path, args.scenario_name, "Matrices")
         if not os.path.exists(mock_result_path):
-            msg = "Mock Results directory {} does not exist.".format(
-                mock_result_path)
-            log.error(msg)
-            raise NameError(msg)
+            log.info("Matrices folder does not exist, copying the folder from test scenario")
+            test_path = os.path.join(args.results_path,"test","Matrices")
+            import shutil
```
zptro, line 64: Imports should be at the top of the module.

zptro (commented):
So to make it clear, this will save you from hassle when using test data as you will not have to change the Results folder in the project settings to make model runs with test data work? But what if you want to use `MockAssignment` with the full helmet data, but you do not have the matrices in the right directory? The model run will then happily continue, but you will then run into a problem of too small impedance matrices at a later stage.

hsl-petrhaj:
> So to make it clear, this will save you from hassle when using test data as you will not have to change the Results folder in the project settings to make model runs with test data work? But what if you want to use `MockAssignment` with the full helmet data, but you do not have the matrices in the right directory? The model run will then happily continue, but you will then run into a problem of too small impedance matrices at a later stage.

Yes, that was the idea, but indeed this does not work well with mock run of real Helmet data. I have to rethink this a bit.

johpiip:
I removed myself from reviewers as this is still in draft phase. Please tag me if you have any questions or we need to discuss something, or mark me as reviewer when this is ready for review. 😊

PR #532, opened 2023-07-04
Feat/noise prediction

s-hiitola:
This should not be merged into this version of Helmet.
Updated noise calculation method based on the last update on the Nordic noise prediction method, as described in 

https://books.googleusercontent.com/books/content?req=AKW5Qaen8XuI4Zzpi1yL840L0unyqTZfBK2-PBLlgS6UJwZpF7uCljYSA8sFP9-pFDwwYLEpPUmY5IOI59S8diml4cVcNUP_z-yANVY9sTdWdB09ITLJZxl518twmqELqP2a8VWIK8u-hHKK_bbpB8axGQFOZ65OUXXDroPlgrcvWr62f9gj6MIa9x8tNC5ceOpQQrXJnb-0ER0JOR9BLKIQxD300-Rtjk9Z6fzJF0-ACc38cQU-SR9EQYsZlMsD9HsiZt5N6Vqmju3kG5auHksVI8z3DhygIg 

pp. 98 - 99

Compared to the currently used method, the areas affected by noise are greatly reduced. The last update included reflections from vertical surfaces, which may make up for the difference when doing actual noise assessments. However, since our model is only a noise emission model with atmospheric attenuation, reflections are not taken into account.

### DIFF
--- a/Scripts/assignment/emme_assignment.py
+++ b/Scripts/assignment/emme_assignment.py
@@ -427,6 +427,7 @@ class EmmeAssignmentModel(AssignmentModel):
             speed = max(speed, 50.0)
             
             # Calculate start noise
+            log.info("NEW FORMULA FOR NOISE CALC")
             if speed >= 50:
                 LAE_light = 73.5 + 25*log10(speed/50)
                 LAE_heavy = 80.5 + 30*log10(speed/50)
@@ -442,6 +443,7 @@ class EmmeAssignmentModel(AssignmentModel):
                 start_noise = 10*log10(10**(LAeq_light/10)+10**(LAeq_heavy/10))
             else:
                 start_noise = 0
+            log.info(start_noise)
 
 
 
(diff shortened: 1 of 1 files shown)


Scripts/assignment/emme_assignment.py
```diff
@@ -425,18 +425,27 @@ def calc_noise(self):
...
+            
             # Calculate start noise
-            if speed <= 90:
-                heavy_correction = (10*log10((1-heavy_share)
-                                    + 500*heavy_share/speed))
+            log.info("NEW FORMULA FOR NOISE CALC")
+            if speed >= 50:
+                LAE_light = 73.5 + 25*log10(speed/50) 
+                LAE_heavy = 80.5 + 30*log10(speed/50)
+            if speed >= 40:
```
zptro, line 434: :exclamation: This is not an `elif` clause, so it will always override the previous `if`.

PR #534, opened 2023-07-05
Network validation vol2

hsl-petrhaj:
I tried to bring all the macro checks to Python, but there are some "strategic" problems. The biggest is that validate_network happens before noboa parameter is assigned to segments. That can only be solved by some separate check, perhaps something like validata_network_after_load could do that. Still work in progress (WIP).


Scripts/utils/network.py
```diff
@@ -0,0 +1,45 @@
...
+                segment.allow_alightings = is_stop
+                segment.allow_boardings = is_stop
+        elif line.mode.id in ['m','r']:
+            for segment in line.segments():
+                is_stop = segment.data1 > 0
+                segment.allow_alightings = is_stop
+                segment.allow_boardings = is_stop
+                print(segment.id, is_stop)
+
+    return network
```
zptro, line 45: I added functionality to mock_project.py that reads stop patterns. As I understand it, this function is thereby obsolete.

zptro (changes requested):
I haven't investigated all the tests in detail, but overall this looks good. I have one suggested change:

johpiip (commented):
Thank you for the PR! 😊 I have two questions:

1. Which validations does this PR add? Based on added messages I think a) checking that vdf can not be zero and b) checking segment speeds before boardings and alightnings. Consider updating the PR description.
2. I tried running this branch on the small test network. I got an error "[ERROR] Link VDF missing for link 197038-197068" which is correct. However, I think vdfs are updated in `_set_bike_vdfs()` and `_set_car_and_transit_vdfs()` based on link type. So, if the validation is done before model run, zero vdfs will give errors even if they will be correctly handled when the model run starts. I am not sure whether we need this vdf check in network validation.

hsl-petrhaj:
> Thank you for the PR! 😊 I have two questions:
> 
>     1. Which validations does this PR add? Based on added messages I think a) checking that vdf can not be zero and b) checking segment speeds before boardings and alightnings. Consider updating the PR description.
> 
>     2. I tried running this branch on the small test network. I got an error "[ERROR] Link VDF missing for link 197038-197068" which is correct. However, I think vdfs are updated in `_set_bike_vdfs()` and `_set_car_and_transit_vdfs()` based on link type. So, if the validation is done before model run, zero vdfs will give errors even if they will be correctly handled when the model run starts. I am not sure whether we need this vdf check in network validation.

1. I added these validation to network validation test:
- Segment check, train or metro travel time us1=0 before stopping (noalin=0 or noboan=0)
- Segment check, train or metro travel time us1 is not 0 before stopping (noalin=1 and noboan=1)
- Line check, headway should not be 0,1
They all come from Mervi's macro, and are covering the parts that were not implemented in the previous pull request.
2. I will look a bit deeper into this, but there was the issue that many checks required loaded network at minimum. I tried to code some of this functionality separately.

hsl-petrhaj:
I removed the VDF checks. Please check again if you think this is ready to merge. The us1 attribute for rail and metro is quite critical as it bit me some weeks ago when creating another scenario. Speed of light transit was way too attractive.
@zptro  @johpiip

Scripts/utils/validate_network.py
```diff
@@ -95,7 +95,7 @@ def validate(network, fares=None):
                 log.error(msg)
                 raise ValueError(msg)
         if network.mode('t') in link.modes or network.mode('p') in link.modes:
-            speedstr = str(int(link.data1))
+            speedstr = str(int(link.data1)).zfill(6)
             speed = {
                 "aht": int(speedstr[:-4]),
                 "pt": int(speedstr[-4:-2]),
```
zptro, line 101: The reason why `speedstr` is sliced backwards from the end is that the string may be either 5 or 6 digits long. `zfill` is an elegant solution to this that would allow us to do the slicing from start to end.

zptro (approved):
Again, I haven't checked this PR in detail, but overall it looks good!

Scripts/utils/validate_network.py
hsl-petrhaj: Sorry, not sure I got the point. Do you mean this implementation is now good? Or are you still proposing some changes?

hsl-petrhaj: @zptro Do you think this branch is ready for merge now?

zptro: Sorry, I had missed this conversation. I mean that the slicing on rows 100-102 could be done in in a way that is easier to understand (without negative values) now that you have performed `zfill`. Otherwise there is no point in doing `zfill`.

zptro (approved):
I think this looks good, so I approve it. But please do not interpret my approval as "this can safely be merged", as that is not something I have checked in detail.

PR #536, opened 2023-07-11
Run uncongested transit in last iteration with flag option in dev_config.json

hsl-petrhaj:
This enables using DO_NOT_USE_CONGESTED_PT flag


hsl-petrhaj:
Probably to be discarded since the same effect can be achieved with vehicle transaction set to values like 99999.

hsl-petrhaj:
Closing this as the method with 99999 per vehicle capacity is much easier.

PR #537, opened 2023-07-12
Datatypes added to the core

hsl-petrhaj:
The point of this pull request is to add data types in so many places that the IDE can figure out the rest. I hope to test the overall consistency with something like mypy. Let's see how it goes.


hsl-petrhaj:
Checked mypy runs:
python -m mypy helmet_validate_inputfiles.py
python -m mypy helmet.py

johpiip (approved):
Looks good to me! :) Model results stay the same when I compared small test model run of `olusanya` to this branch.

hsl-petrhaj:
More validation of data types would be good to implement like:
`self.assertIs(type(impedances), dict)`
Otherwise there is theoretical chance the mypy types are not equal to the real Python types.
That can be done if some issue related to types is discovered.

PR #538, opened 2023-08-07
Point error message to zone_mapping file if appropriate

zptro:
Until now, the error message has always pointed to the data file, even if the zone number error is actually in the zone mapping file. This PR fixes that, so that the error message points to the zone mapping file if the error is to be found there. If the error is indeed in the data file, the zone number is transformed using the zone mapping when applicable.


johpiip (approved):
Good fix!

PR #539, opened 2023-08-07
Refactor logit models

zptro:
This PR changes the the way the *mode choices* are done in agent simulations. Instead of Monte Carlo choices, the choices are based on direct utility maximization including an individual error term. The logic behind this is described in `EXT-MAL 2023/Vaikutusten arviointi/Menetelmäkehitys/agenttimallinnus`. Destination choices are still done with Monte Carlo simulation, because direct utility maximization is slow with so many alternatives to compare.

The pros of this change:
- Mode choice will become consistent on the agent level: If the accessibility of a mode increases, agents will switch from other modes to that mode, and no other mode switches will occur.
- Individual dummies will be easy to add to mode choice with minor implications for run time.


johpiip (commented):
Should this change affect results in both agent modelling and normal modelling? I ran the small test network with `olusanya` branch and this branch using normal model run and the results were different. Is it on purpose or by accident?

zptro:
> Should this change affect results in both agent modelling and normal modelling? I ran the small test network with `olusanya` branch and this branch using normal model run and the results were different. Is it on purpose or by accident?

That is strange. Yes, the calculation of the log variables is changed, but it gives the same results according to the integration tests. 😮

johpiip:
Yes, it is! in `result_summary.txt`, assigned demand and mode share is the same in both runs but vehicle kilometres change as well as some accessibility measures. The network is the same. The changes are strangely large:
kuva

johpiip:
Could this branch be slightly out-of-date compared to `olusanya`? So that the differences actually come from other commits than these ones? I tested earlier today that `v4.1.3` and `olusanya` give different results and started an issue #540 if someone could identify which PRs create those changes.

hsl-petrhaj (approved):
Sorry for the delay.

hsl-petrhaj:
@zptro This commit did actually change the leisure demand matrices significantly. I am wondering if we should check this more thoroughly.

PR #548, opened 2023-10-20
Fix critical bug in agent simulation secondary destination choice

zptro:


hsl-petrhaj (approved):
Interesting bug, interesting that Python did not crash with this. Would be good to test if the results are the same as before.

PR #549, opened 2023-11-03
Restore aux modes in transit trip part matrix calculation

### LINKED ISSUE #541: Modes in transit cost calculation?
https://github.com/HSLdevcom/helmet-model-system/blob/5a14cc7d5de89f529a191ba2076f2cf91d59bfef/Scripts/assignment/datatypes/transit.py#L89

This might be a problem for analyzing `aux` transit time matrix.

zptro:
Fix #541


zptro:
> Not sure why is this needed, but it also does not seem to break anything, so why not.

We calculate aux travel time mtrices using the modes in this set. If walk mode is not included, the matrix will contain nothing.

hsl-petrhaj:
Thanks that helps.

PR #553, opened 2023-11-22
Adding the possibility to check the origin, destination and secondary destination for all agent journeys

hsl-petrhaj:
The only change here is that tours.txt gets additional informational about all the important zones. Ideally that would allow for some quick filtering by origin and destination. If we could additionally manage to print the route as well it might be possible to do traversal analysis outside of Emme. Sorry if this is getting too Matsim.
I also know that for example the mode split of the agent-based model should still be validated, so it is just building on top of the experimental branch.


hsl-petrhaj:
I changed it to draft because there is this strange bug with the zone number. It should not be hotfixed like that, but probably needs deeper focus to figure it out. But maybe some of you knows immediately why it is like that.

PR #558, opened 2024-01-08
Update README.md

zptro:
`%APPDATA%\Python\Scripts` has at some point changed to `%APPDATA%\Python\Python37\Scripts`. This change has no impact if Emme-bundled python is used.


hsl-petrhaj (approved):
I checked my paths, seems correct, so I approve.

PR #562, opened 2024-02-22
adjusted the links.txt printing

hsl-petrhaj:
This should simplify some analyses done with links.txt file.


zptro (approved):
Some line breaks in the code could be nice, but otherwise this looks good!

PR #565, opened 2024-03-18
Update VS Code instructions

zptro:
Solve issue with testing in updated version of VS Code

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,9 +157,13 @@ Alternatively, you can run the mock assignment with the help of command line
 
 The following extensions are recommended when developing with Visual Studio Code:
 
-- [Python extension version<2024.0.0](https://marketplace.visualstudio.com/items?itemName=ms-python.python)
+- [Python extension v2023.25.XXX](https://marketplace.visualstudio.com/items?itemName=ms-python.python)
   (version >=2024.0.0 comes with pytest 8, which does not support python 3.7)
 - [Python Debugger v2024.0.XXX](https://marketplace.visualstudio.com/items?itemName=ms-python.debugpy)
+
+Alternatively, these test views seem to work with any version of the Python and Python Debugger extensions:
+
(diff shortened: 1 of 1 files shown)


hsl-petrhaj:
@zptro I did not understand from the instructions, if this also works in Python 3.7. Which Python version have you used?

zptro:
> @zptro I did not understand from the instructions, if this also works in Python 3.7. Which Python version have you used?

Python 3.7 is what I use and what did no longer work after my VS Code extensions were updated. Reverting these updates fixed the issue.

hsl-petrhaj:
Ok, if it works that way, I will accept it now. For me pytest kept working somehow though.

hsl-petrhaj (approved):

zptro:
> Ok, if it works that way, I will accept it now. For me pytest kept working somehow though.

Oh, can you see the tests in the Testing explorer in VS Code even if you have Python extension version 2024.0.0 or above?

hsl-petrhaj:
I have the following:
Python 3.7
Test Explorer UI v2.21.1
Python Debugger 2024.2.0

Seems to work fine except for some logging error.

zptro:
> I have the following: Python 3.7 Test Explorer UI v2.21.1 Python Debugger 2024.2.0
> 
> Seems to work fine except for some logging error.

I see. I haven't been using Test Explorer UI lately, as it is described as deprecated, and Python extension's test view has worked just as well. But I guess that the benefit of being deprecated is that it still works well with Python 3.7. :)

zptro:
@hsl-petrhaj Could you look at this so we can close it?

hsl-petrhaj:
Yes, I can check this.

hsl-petrhaj (approved):
Seems like I was running with a slightly older version before, now I bumped into the same issue and your fix indeed works.

samakinen (approved):

PR #567, opened 2024-06-17
Fix/emme 24 compatibility

samakinen:
OpenPaths Emme 24 has been released with Python 3.11 and updated Numpy and Pandas versions. The current Helmet model system is not compatible with the new versions.
This PR will make the model system compatible with Emme 24 while maintaining compatibility with Emme 23 (tested) and probably other versions using Python 3.7 (not tested).
While testing I noticed some differences between different Emme and Helmet branches. I run the same network and land use scenarios for 3 different test cases:
1. Emme 23, olusanya
2. Emme 23, fix/emme 24 compatibility
3. Emme 24, fix/emme 24 compatibility
I run CBA calculation to compare the calculation results and found 3-6 M€/a differences between all scenarios caused by travel time saving. The code changes and the Emme version should not change the calculation results and I could not find any logical differences between the 3 scenarios. It seams the differences are caused by fluctuations in the demand model.
I marked the PR as a draft until the reason for differences can be identified.

### DIFF
--- a/Scripts/demand/trips.py
+++ b/Scripts/demand/trips.py
@@ -106,7 +106,7 @@ class DemandModel:
         numpy.random.seed(param.population_draw)
         self.population = []
         zone_numbers = self.zone_data.zone_numbers[self.bounds]
-        self.zone_population = pandas.Series(0.0, zone_numbers)
+        self.zone_population = pandas.Series(0, zone_numbers)
         # Group -1 is under-7-year-olds
         age_range = numpy.arange(-1, len(param.age_groups))
         for zone_number in zone_numbers:
--- a/Scripts/demand/trips.py
+++ b/Scripts/demand/trips.py
@@ -106,7 +106,7 @@ class DemandModel:
         numpy.random.seed(param.population_draw)
         self.population = []
         zone_numbers = self.zone_data.zone_numbers[self.bounds]
-        self.zone_population = pandas.Series(0, zone_numbers)
+        self.zone_population = pandas.Series(0.0, zone_numbers)
         # Group -1 is under-7-year-olds
         age_range = numpy.arange(-1, len(param.age_groups))
         for zone_number in zone_numbers:
--- a/Scripts/models/generation.py
+++ b/Scripts/models/generation.py
@@ -26,7 +26,7 @@ class GenerationModel:
 
     def init_tours(self):
         """Initialize `tours` vector to 0."""
-        self.tours = pandas.Series(0, self.purpose.zone_numbers)
+        self.tours = pandas.Series(0.0, self.purpose.zone_numbers)
 
     def add_tours(self):
         """Generate and add (peripheral) tours to zone vector."""
--- a/Scripts/utils/read_csv_file.py
+++ b/Scripts/utils/read_csv_file.py
@@ -1,9 +1,7 @@
-from decimal import DivisionByZero
-from itertools import groupby
 import os
 from typing import Optional
 import pandas
-import numpy # type: ignore
+import numpy
 
 import utils.log as log
 
@@ -49,9 +47,11 @@ def read_csv_file(data_dir: str,
         raise NameError(msg)
     header: Optional[str] = None if squeeze else "infer"
     data: pandas.DataFrame = pandas.read_csv(
-        path, delim_whitespace=True, squeeze=squeeze, keep_default_na=False,
+        path, sep='\s+', keep_default_na=False,
         na_values="", comment='#', header=header)
-    if data.index.is_numeric() and data.index.hasnans: # type: ignore
+    if squeeze:
+        data = data.squeeze()
+    if pandas.api.types.is_numeric_dtype(data.index) and data.index.hasnans:
         msg = "Row with only spaces or tabs in file {}".format(path)
         log.error(msg)
         raise IndexError(msg)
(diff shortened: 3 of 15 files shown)


hsl-petrhaj:
Hi Sami, unable to check these in detail now, but could you add the new pandas version to the requirements? Could there be some differences because of that?

samakinen:
Pandas is not listed in the requirements. Only requirements is openpyxl that had to be updated from 2.6.4 to 3.1.4 when running with python 3.11. When using old Emme with Python 3.7 the old version will be used. Pandas comes directly from Emme environment and the model system has to use the version provided. This is 0.24.2 when using Emme 23 and 2.0.2 when using Emme 24. I don't think Pandas version would (or should) explain the differences because my two test cases "emme 23, olusanya" and "emme 23, fix/emme-24-compatibility" used the same Pandas version but still produced different results.

Scripts/utils/read_csv_file.py
```diff
@@ -68,17 +68,17 @@ def read_csv_file(data_dir: str,
...
         map_path = os.path.join(data_dir, "zone_mapping.txt")
         if os.path.exists(map_path):
             log_path = map_path
-            mapping = pandas.read_csv(map_path, delim_whitespace=True).squeeze()
+            mapping = pandas.read_csv(map_path, sep='\s+').squeeze()
             if "total" in data.columns:
                 # If file contains total and shares of total,
                 # shares are aggregated as averages with total as weight
-                data = data.groupby(mapping).agg(avg, weights=data["total"])
+                data = data.groupby(mapping).agg(lambda ser: avg(ser, weights=data["total"]))
```
zptro, line 81: What was the problem here? I see no indications in the pandas documentation that feeding regular functions into `agg` would have been deprecated.

Scripts/utils/zone_interval.py
```diff
@@ -183,7 +183,7 @@ def add(self, orig, dest):
         dest : int
             Tour destination zone number
         """
-        self.matrix.at[self.mapping[orig], self.mapping[dest]] += 1
+        self.matrix.at[self.mapping[orig], self.mapping[dest]] += 1.0
```
zptro, line 186: I do not understand why integers cannot be handled as integers?

zptro:
> Pandas is not listed in the requirements. Only requirements is openpyxl that had to be updated from 2.6.4 to 3.1.4 when running with python 3.11. When using old Emme with Python 3.7 the old version will be used. Pandas comes directly from Emme environment and the model system has to use the version provided. This is 0.24.2 when using Emme 23 and 2.0.2 when using Emme 24. I don't think Pandas version would (or should) explain the differences because my two test cases "emme 23, olusanya" and "emme 23, fix/emme-24-compatibility" used the same Pandas version but still produced different results.

True, pandas and most other packages is not in requirements.txt. But pipfile needs to be updated, so that unit tests behave in the same way as production version of the model system.

Scripts/demand/trips.py
```diff
@@ -106,7 +106,7 @@ def create_population(self):
         numpy.random.seed(param.population_draw)
         self.population = []
         zone_numbers = self.zone_data.zone_numbers[self.bounds]
-        self.zone_population = pandas.Series(0, zone_numbers)
+        self.zone_population = pandas.Series(0.0, zone_numbers)
```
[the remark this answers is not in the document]
samakinen: Good catch, this one should be integer. I think floats would work as well, but int is better if the population is an integer value.

Scripts/utils/read_csv_file.py
samakinen: I could not make the avg function work without this. It seems that avg function expects to get a Series as an argument, but will get a DataFrame object instead. For some reason adding lambda here fixes that. There might be a more elegant way to fix this.

Scripts/utils/zone_interval.py
samakinen: self.matrix can not be integers here because we will add floating point values to it in other methods. Adding 1 instead of 1.0 here would work as well (implicit conversion from int to float), but I think it's more clear to be consistent with the datatypes.

samakinen:
> True, pandas and most other packages is not in requirements.txt. But pipfile needs to be updated, so that unit tests behave in the same way as production version of the model system.
Are we ready to switch the "production version" to be the one defined by Emme 24? I run the unit tests in separate virtualenvs for Emme 23 and Emme 24. I can update the pipfile to the new package versions if we are ready to switch.

zptro:
> > True, pandas and most other packages is not in requirements.txt. But pipfile needs to be updated, so that unit tests behave in the same way as production version of the model system.
> > Are we ready to switch the "production version" to be the one defined by Emme 24? I run the unit tests in separate virtualenvs for Emme 23 and Emme 24. I can update the pipfile to the new package versions if we are ready to switch.

I cannot say what is the right solution for Helmet version "bleeding edge", but I think we could schedule the migration of LEM to the upcoming weeks.

Scripts/utils/zone_interval.py
zptro: The agent model and the aggregate model behave in very different ways here, the agent model is using `add` and aggregate model `aggregate`. One idea would be to separate them better, keeping integers in the agent version.

hsl-petrhaj:
Helmet 5 at least should be switched to Emme24. Can the compatibility with Emme23 be retained though? Or is there something that would break the compatibility?

Scripts/models/generation.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, purpose, resultdata):
 
     def init_tours(self):
         """Initialize `tours` vector to 0."""
-        self.tours = pandas.Series(0, self.purpose.zone_numbers)
+        self.tours = pandas.Series(0.0, self.purpose.zone_numbers)
 
     def add_tours(self):
         """Generate and add (peripheral) tours to zone vector."""
```
zptro, line 32: This may have small impacts on the demand, because I think in the base case the `self.tours` vector will change to `float32` (because that is what is added from `self.zone_data`) in `add_tours`. Now when `self.tours` is implicitly initialized as `float64`, this will probably be broadcasted to large parts of the demand model. I suggest adding a `dtype=numpy.float32` to see if that changes results. This could also be tested in `ExternalModel`.

samakinen: My understanding is that Pandas does not change the Series datatype on addition. In the earlier version the series was created with int64 dtype and each addition of real values would only add the integer part of the number (`floor()`). Comparing to that the difference between float32 and float64 should be minimal. We can add `dtype=numpy.float32` here if want to reduce memory consumption, especially if there are bigger matrices calculated based on this Series and same dtype.

Scripts/utils/zone_interval.py
samakinen: Is there a good reason to duplicate code because of it? As far as I can see, adding 1 to float64 or int64 values works identically up to 2<sup>53</sup> (or 2<sup>24</sup> for float32). Those values are probably more than we need in this case. If the number range could be an issue in some cases we should use float64.

samakinen:
I made a new commit to update the development environment and documentation to refer OpenPaths EMME 24 and python 3.11. I also updated the github actions to run the tests with the new dependencies and python version. The model-system will still works with older EMME and Python 3.7, but there are no automated tests to make sure it will do so in the future.

Scripts/models/generation.py
zptro: We discussed this and it turned out that we were both wrong. `add_tours` has always changed `self.tours` into float64.

zptro (approved):
This looks good now, I think! However, I have no idea why the results are fluctuating.

PR #573, opened 2024-09-10
Disable turn restrictions for bikes

jalovaj2:
Sets turn penalties zero for bike mode in the traffic assignment.


jalovaj2:
The Network objects used in the tests don't have Turn objects defines so have to modify the tests to include them

PR #575, opened 2024-10-07
Feat/park and ride model estimation

hsl-petrhaj:
Here is a proposal for implementing the park and ride assignment logic. I think we could merge this and then the for loop inside the split_park_and_ride function can be rewritten in pure numpy. 
Tested with updated test network.


hsl-petrhaj:
Everything seems to be merged into helmet5-experimental

PR #577, opened 2024-10-23
Fix revenues summed too many times

samulikyytsonen:
This affects only to the results in .txt file - not CBA-excel


zptro (dismissed):
Nice catch!

hsl-petrhaj:
@samulikyytsonen I am not so familiar with this part, but why is the proposed behavior inconsistent with how cars are counted? Is there some underlying difference?

hsl-petrhaj (approved):
Now it looks consistent, thank you for the PR. You can squash merge that into olusanya now (as usual).

samulikyytsonen:
Seems that I don't have rights to do the merge. I need to be authorized for that

hsl-petrhaj:
Thanks for letting me know, it is merged now. I will check the policies later in more detail.

PR #578, opened 2024-10-23
Feat/bike hills modeling

hsl-petrhaj:
Key improvements for bike routing in Helmet model. Mostly based on master's thesis.


hsl-petrhaj:
Now merged into helmet5-experimental branch.

PR #580, opened 2025-02-06
Feat/model system events

samakinen:
Add possibility to attach external modules to the model system. These modules are called at various parts of the model run. New modules can be added by adding a .py file containing one or more implementation of classes inheriting ModelSystemListener class into Scripts/events/examples (to be changed) directory.


hsl-petrhaj:
I merged this the old way as it looks better on my git tree.

PR #582, opened 2025-04-08
Feat/export estimation data

samakinen:
Add support for exporting zone and network data into gpkg-files.
Add geodata_helpers.py for handling network geodataframe operations.
Split definition of EventHandler into a separate file.


hsl-petrhaj:
Debug in the log file seems to be unnecessary big, could we reduce it?
@samakinen

samakinen:
> Debug in the log file seems to be unnecessary big, could we reduce it? 

@hsl-petrhaj, what kind of debut log do you get? For me there is only one line notification that the module has been loaded.

PR #602, opened 2026-01-16
Validate connectors

s-hiitola:
Added validation of connectors, removed warning when trams are allowed on light rail links

Also fixes a possible crash scenario when calculating noise.
Small corrections to test data (municipality codes and connector link types for pnr centroids)


hsl-petrhaj (approved):
Looks good

PR #614, opened 2026-02-12
Adding the option of automated project creation, based on Valma code,…

hsl-petrhaj:
Adding the option of automated project creation, based on Valma code, added projection, view and basemap.
Dev config bumped from version 5.0.0 to 5.0.2


hsl-petrhaj:
@s-hiitola @johpiip I tagged you just to let you know that this automated workflow is now possible. Next we should add the option to the UI and add the option in helmet-docs.

s-hiitola (approved):
Fine for now. Making it possible to explicitly define the database dimensions instead of using presets could be a good idea.

PR #616, opened 2026-03-04
Refactor validation

s-hiitola:
- Errors will be grouped instead of failing at the first error
- Validation scripts are split up in functions
- Matrix space validation
- Fixed a bug with tram/light rail node validation
- Rewrote scenario numbering during validation, added scenario ids to logs


hsl-petrhaj (approved):
Looks good.

PR #617, opened 2026-03-05
Feat/modes in links txt

### LINKED ISSUE #607: links.txt should have modes
They would enable easier filtering

s-hiitola:
Added modes column as a string, exactly like in Emme.


hsl-petrhaj:
Fixes #607

PR #621, opened 2026-03-11
Fixing zone data for tests

hsl-petrhaj:
Removing old bugs in zone data for tests.


hsl-petrhaj:
Do not merge yet, it seems like it is actually some ancient feature.

hsl-petrhaj:
https://github.com/HSLdevcom/helmet-model-system/pull/460 This seems to be the feature.

hsl-petrhaj:
Rejecting, because this is a feature, not a bug.

PR #632, opened 2026-04-27
Helmet 5.0.2

hsl-petrhaj:
* Transit zone tool for toolbox (change D zone to E zone)
* Select scenario for link printing in toolbox
* Print modes as string in links.txt
* Major results refactor - results are now event-based
* Automatic project generation (based on Valma's script)
* Validation refactor
* Aggregated P&R model for agent-based modelling


s-hiitola (approved):
All good

PR #637, opened 2026-06-01
Helmet51

hsl-petrhaj:
Expected changes in Helmet 5.1:

- Expansion factors updated to correct values
- CBA updated to 2023 and new expansion factors
- Volume delay function updated in part of network
- Enables more detailed TAZs
- More options for public transport ticket modelling


hsl-petrhaj:
Should be merged into dev not main

PR #647, opened 2026-08-03
Add scenario number to the name of the matrix

### LINKED ISSUE #646: Emme Matrix does not have any reference to the scenario
It would be good if matrices can be associated to scenarios that produced them.

hsl-petrhaj:
Fixes #646


hsl-petrhaj:
Close #646

PR #652, opened 2026-08-07
Testing network connectivity with Emme assignment.

hsl-petrhaj:
Here is my version for testing the connectivity of basic four nodes. Can be improved with optional flag and testing cargo modes as well. 
Island centroids will need some care as well, right now the testing only accepts Suomenlinna.

### DIFF
--- a/Scripts/utils/validate_network.py
+++ b/Scripts/utils/validate_network.py
@@ -424,7 +424,27 @@ def validate_network_connectivity(modeller, scenario):
     EXTERNAL_RAILWAY_CENTROIDS = [z for z in zone_numbers.values() if z in set(range(34300, 34400))]
     #Make simple assignment to get impedance matrices
     mf1 = modeller.emmebank.matrix("mf1").get_numpy_data(scenario_id=scenario.id)
-    log.info(str(mf1))
+
+    test_func = 50
+    for idx in [f"fd{test_func}", f"ft{test_func}", f"fp{test_func}"]:
+        try:
+            modeller.emmebank.delete_function(idx)
+        except Exception:
+            pass
+        modeller.emmebank.create_function(
+            idx, "1")
+
+    network = scenario.get_network()
+    for link in network.links():
+        link.volume_delay_func = test_func
+        link.num_lanes = 1
+    for segment in network.transit_segments():
+        segment.transit_time_func = test_func
+    for turn in network.turns():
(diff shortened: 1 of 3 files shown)


hsl-petrhaj:
TODO: Needs some handling handling of vdfs and extra parameters. There are differences between empty scenario and already ran network.

Scripts/utils/validate_network.py
```diff
@@ -285,3 +287,218 @@ def validate_transit(network):
...
+    #Zone numbers
+    zone_numbers = {zone: i for i, zone in enumerate(scenario.zone_numbers)}
+    modes = ["car","bike","walk","transit"]
+    assignment_methods = {
+        "car": modeller.tool("inro.emme.traffic_assignment.sola_traffic_assignment"),
+        "bike": modeller.tool("inro.emme.traffic_assignment.standard_traffic_assignment"),
+        "walk": modeller.tool("inro.emme.transit_assignment.standard_transit_assignment"),
+        "transit": modeller.tool("inro.emme.transit_assignment.standard_transit_assignment")
+    }
+    emme_specs = {
```
s-hiitola, line 317: Could the specs be stored in their own file, such as utils/validation_data/emme_specs?

Scripts/utils/validate_network.py
```diff
@@ -285,3 +287,218 @@ def validate_transit(network):
...
+            },
+            "aux_transit_time": {
+                "perception_factor": 1
+            },
+            "od_results": {
+                "transit_times": "mf2"
+            },
+        },
+        }
+    Suomenlinna = 1531 #TODO: Accept other island centroids as well
```
s-hiitola, line 421: A dictionary with zone titles and numbers might be a more elegant data structure if the number of exceptions that need handling grows

s-hiitola (commented):
Take a look at the comments before merging

Scripts/utils/validate_network.py
hsl-petrhaj: I would keep it concise for now, at least as long as other specs are not moved to similar place.

Scripts/utils/validate_network.py
hsl-petrhaj: Will think about that.

s-hiitola (approved):
Okay to merge as long as it's tested on a full network

PR #654, opened 2026-08-13
Fix/cba raise error for missing vehicles

s-hiitola:
CBA run will now halt if there are vehicles with non-zero mileage that are ignored by CBA script.

Ferries or "lautta" is whitelisted and will not be taken into account, as it has not been. This should be fine for most scenarios, but might cause problems later if there are scenarios with major increases in HSL ferry traffic. Changes in demand will be calculated correctly, but operational costs are ignored.


hsl-petrhaj (approved):
Looks good.

PR #657, opened 2026-08-17
Changing walking assignment to extended

hsl-petrhaj:
We had lots of troubles with standard transit assignment. When we check the speed as 60*mf26/mf16 we got any speeds from 5,02 to 5,4 km/h. This is unacceptable. Extended transit assignment does not seem to have the same problem.
@zptro FYI.

Changing walking assignment to extended, moving bike assignment to is own file.


s-hiitola (approved):
All good

